PR #73, opened 2018-06-25
Fix issue where plugin was published with -dirty tag

### LINKED ISSUE #66: Plugins published with dirty tag

ellismg:
Our build was creating a dirty work tree, because of changes to the
sdk/python/pulumi_azure.egg-info directory. In other projects, we
didn't run into this, so I synced our Makefile and build/common.mk
files to match what we had in pulumi-aws.

Now, the build does not dirty the worktree and we can delete this
checked in sdk/python/pulumi_azure.egg-info folder.

As a belt and suspenders measure, I changed the way the
publish-plugin.sh computes the version. We now use the exact version
that is in the package.json for the node package, which is the exact
version that we'll try to install. That alone would have fixed the
issue but it is nice to bring this stuff in line with the other
repositories and not dirty the worktree while building.

Fixes #66


ellismg:
@swgillespie I'm going to merge this as is, but let me know if you have any follow-up feedback.

swgillespie:
ack, I'm sorry I completely missed this! I'll do a quick review now. I need to get my github notifications in order...

swgillespie:
LGTM 👍

PR #104, opened 2018-09-13
Remove the dns name component of this test.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-azure-serverless/issues/19

The test doesn't really need this property set.  And setting it causes tests to have issues in parallel since two simultaneous runs of this test will both try to take the same dns-name-label which Azure doesn't allow.


ellismg:
LGTM

PR #106, opened 2018-09-14
Update name maxlength for servicehub namespaces.

CyrusNajmabadi:


lukehoban:
As pointed out to me when I quoted another PR with the term in it's title - there is no such thing as "servicehub namespaces".  There are [Azure Event Hubs](https://azure.microsoft.com/en-us/services/event-hubs/) and [Azure Service Bus](https://azure.microsoft.com/en-us/services/service-bus/).

PR #114, opened 2018-09-17
Bump the referenced TF version.

pgavlin:
THe version of TF we're locked to is quite old, and does not play well
with some of our other dependencies. This is a provisional approach to
see how things work with a newer version of TF.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -352,6 +352,7 @@
     "helper/schema",
     "helper/structure",
     "helper/validation",
+    "httpclient",
     "moduledeps",
     "plugin/discovery",
     "registry",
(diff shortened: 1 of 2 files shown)


pgavlin:
@lukehoban I'm running some of the `pulumi-cloud` tests locally using this build. Things seem to be working.

pgavlin:
> I suppose the only reason we do this currently is because pulumi-Terraform brings in a constraint to a different version and this is how we break the tie. But this is quite brittle, and will likely lead to us accidentaly using the wrong version in the future.

Yes, I agree.

> Is there a way to set a looser constraint in Pulumi-Terraform such that we can solve correctly without need for this override at all?

There may be, yes. I _think_ that we'd need to set the version in PTF to the minimum version referenced by the various TF providers and then hope that no breaking changes have been made that prevent the bridge code from compiling. In the limit we could solve this with a branch of PTF per major TF version if necessary.

PR #125, opened 2018-09-26
Update Gopkg.toml using govendor-override.

pgavlin:
These changes add explicit overrides for the transitive dependencies of
`terraform-provider-azurerm` at the versions specified in that project's
`vendor.json` file. This ensures that we will build using package
versions with which the TF provider is known to work.

These changes also update the referenced version of the TF provider.


Gopkg.toml
```diff
@@ -3,23 +3,332 @@
...
-  name = "github.com/pulumi/pulumi-terraform"
   branch = "master"
+  name = "github.com/pulumi/pulumi-terraform"
 
 [[constraint]]
+  branch = "pulumi-master"
   name = "github.com/terraform-providers/terraform-provider-azurerm"
   source = "github.com/pulumi/terraform-provider-azurerm"
-  branch = "pulumi-master"
+
```
lukehoban: Is the newline expected here - it's looks surprising to my eyes.

lukehoban:
BTW - What version of the TF provider are we now delivering here?  Would love to be clear about that in PR description/title any time will pull in new bits.

pgavlin:
> What version of the TF provider are we now delivering here? Would love to be clear about that in PR description/title any time will pull in new bits.

1.15.0. I will add that to the commit message.

CyrusNajmabadi:
Thanks.  You rock!

PR #126, opened 2018-10-01
Don't auto-name when 'name' is optional

### LINKED ISSUE #121: Azure role assignment name needs to be GUID
When trying to assign a role to a service principal, the name needs to be a GUID.

The following code -
```
 let roleAssignment = new azure.role.assignment("ra", {
      principalId: servicePrincipal.id,
      roleDefinitionName: "Contributor",
      scope: `/subscriptions/${azure.config.subscriptionId}`
})
```

Will return a 400 error from azure saying that the name must be a valid GUID.

I tested with the `uuid` package and it worked, however this will trigger a change every time `pulumi up` is ran.

```
import * as uuid from "uuid"
let role = new azure.role.assignment("ra", {
     name: uuid.v4(),
     principalId: _sp.id,
     roleDefinitionName: "Contributor",
     scope: `/subscriptions/${azure.config.subscriptionId}`
})
```

lukehoban:
We have been applying auto-naming to all input properties, even if they were already Optional.  In that case, the provider already understands how to provide a default name - or one is truly not required at all.  So we should avoid overwriting this with a Pulumi autoname.

Also fixes the casing of `azure.role.assignment` - this is a breaking change.

Fixes #121.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -209,7 +209,7 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_autoscale_setting": {Tok: azureResource(azureAutoscale, "Setting")},
 
 			// Authorization
-			"azurerm_role_assignment": {Tok: azureResource(azureRole, "assignment")},
+			"azurerm_role_assignment": {Tok: azureResource(azureRole, "Assignment")},
 			"azurerm_role_definition": {Tok: azureResource(azureRole, "Definition")},
 
 			// Azure Container Service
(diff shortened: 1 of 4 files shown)


lukehoban:
@pgavlin If this looks good, we should make the same change in all other TF providers.

lukehoban:
In the case of #121, the TF provider is autonaming. Critically, it is providing a GUID as the autoname, which is the only allowed value. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_role_assignment.go#L94

We could in theory replicate the same logic here (and in however many other places similar thing exist). But it generally seems wrong to override Optional inputs by default - there are a few reasons the name may have been marked as optional - and most suggest we should not be automatically providing a new name ourselves.

lukehoban:
Likely worth seeing how many properties this impacts currently though - I’ll pull that list.

joeduffy:
In many cases, TF will just generate a random string of digits, unconnected to the name in the Pulumi program. Sometimes with a "terraform-" prefix. This seems less appealing than a meaningful name that correlates back with your program. I get why the GUID case breaks as a result, but this doesn't seem like the majority case.

We should probably step back and think deeper about the design here, since there are subtle and important implications of any fundamental change, and only make targeted fixes until then.

lukehoban:
Looked a little more, and @joeduffy is right that this would have a broader impact than intended in the AWS provider, where many resources do some form of their own autonaming with names that would be less ideal.

In Azure, the only place this changes anything is the specific property that triggers #121 - so we'll just patch that as a one-off.

Replaced with https://github.com/pulumi/pulumi-azure/pull/127.

PR #127, opened 2018-10-01
Don't autoname azure.role.Assignment

lukehoban:
This resource is autonamed to a GUID in the underlying provider - which is the only allowed value - and so should not be autonamed in the Pulumi layer.

Also fixed capitalization from `assignment` to `Assignment` - this is a breaking change.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -209,7 +209,13 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_autoscale_setting": {Tok: azureResource(azureAutoscale, "Setting")},
 
 			// Authorization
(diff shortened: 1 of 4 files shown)


resources.go
```diff
@@ -209,7 +209,13 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_autoscale_setting": {Tok: azureResource(azureAutoscale, "Setting")},
 
 			// Authorization
-			"azurerm_role_assignment": {Tok: azureResource(azureRole, "assignment")},
+			"azurerm_role_assignment": {
+				Tok: azureResource(azureRole, "assignment"),
```
pgavlin: Should this be `Assignment`?

PR #130, opened 2018-10-05
Break out eventgrid, eventhub, and servicebus into their own namespaces.

CyrusNajmabadi:
Note: this should likely not go in until we cut our next release.  This will def be a breaking change.


CyrusNajmabadi:
@lukehoban Please take a look.  Note: there is likely a lot more to do here.  But this wa a fairly sane and simple change.  It matches both terraforms, and Azure's view on things afaict: https://azure.microsoft.com/en-us/blog/events-data-points-and-messages-choosing-the-right-azure-messaging-service-for-your-data/

CyrusNajmabadi:
Ping @lukehoban

CyrusNajmabadi:
@lukehoban Please take a look. Note: there is likely a lot more to do here. But this wa a fairly sane and simple change. It matches both terraform's, and Azure's view on things afaict: https://azure.microsoft.com/en-us/blog/events-data-points-and-messages-choosing-the-right-azure-messaging-service-for-your-data/

CyrusNajmabadi:
Ok.  Moving forward on this.

PR #153, opened 2018-12-02
Update to Terraform Provider v1.19.0

jen20:
This pull request updates `pulumi-azure` to reflect changes made in v1.19.0 of the underlying Terraform AzureRM provider.

There is a breaking change in Kubernetes support - RBAC is now supported in the underlying provider, so our custom addition of `enable_rbac` is removed in favour of the native support. This is documented in the CHANGELOG.


lukehoban:
> Do you know if any of our AKS examples need to change due to the break?

Nothing in our own examples - but I am aware of one customer who will be impacted and I will reach out to them directly as an extra measure.

https://github.com/search?q=org%3Apulumi+enableRbac&type=Code

Strange that the Azure Terraform provider ended up picking a somewhat randomly different name than the actual API: https://docs.microsoft.com/en-us/rest/api/aks/managedclusters/createorupdate#request-body.

PR #161, opened 2018-12-29
Ingest latest Terraform code-gen

joeduffy:
This ingests the latest Terraform code-gen, bringing in the
ability to convert HCL examples to Pulumi in the API docs.

Notes on metrics:

- 461 docs sections ignored
- 22 code blocks failed to convert
- 246 code blocks successfully converted

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -594,7 +594,7 @@
     "pkg/tfbridge",
     "pkg/tfgen"
   ]
-  revision = "2c187fb6f5e33adbc3c54c4dff4c86f7c9af7379"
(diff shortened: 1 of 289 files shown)


joeduffy:
This failed due to a dirty work-tree. I suspect this is because there's a `tf2pulumi` error that causes these examples to get skipped (we silently ignore them and keep going -- in the future, some baselining mechanism is probably a better approach, but at least this caught the problem). I'll have to dig into why this is. It's possibly something about the state of the provider on the build machine.

jen20:
This can be closed now that #164 is merged.

PR #166, opened 2019-01-20
Update Gopkg.{lock,template.toml,toml} for dep 0.5

jen20:
This PR updates us for dep 0.5 for the Azure provider.

The constraints needed modifying in order to allow a solve to complete, however the resulting set of dependencies looks correct (meeting the various minimum version constraints we previously had).

CI will fail since I reverted the script upgrade for 0.5.

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,11 +2,14 @@
 
 
 [[projects]]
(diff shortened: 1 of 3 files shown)


jen20:
Force pushed to rebase onto master following some changes today.

PR #185, opened 2019-02-19
Add link to credentials guide to README

### LINKED ISSUE #170: Write an opinionated guide to how to authenticate for different scenarios
Azure provides many different ways to authenticate, which are appropriate in different circumstances. We should have a document which explains which types of credentials are appropriate in which situation (e.g. Pulumi on a dev machine, in CI, on an Azure VM etc) with specific instructions on how to find the relevant pieces of information.

jen20:
Fixes #170.


ericrudder:
i think you get to use [ or (, but not both

They are enumerated [in the quickstart guide)(

i'd offer just "They are enumerated in the Quickstart" and elide both.  :)

jen20:
Thanks for pointing that out, @ericrudder! I've fixed the markdown syntax to retain the link to make it easier to find the Quickstart guide without resorting to google or site navigation.

PR #201, opened 2019-03-11
Explicitly map name of use32BitWorkerProcess in AppService

### LINKED ISSUE #157: Preview fails when setting FunctionApp/AppService siteConfig.use32BitWorkerProcess property
Preview fails when setting siteConfig.use32BitWorkerProcess property on a FunctionApp or AppService. It looks like the tf mapping is off by a character based on the error message: `error: azure:appservice/appService:AppService resource 'abc-stage-clbdir' has a problem: site_config.0: invalid or unknown key: use32_bit_worker_process`

use32_bit_worker_process looks like it should be passed down as [use_32_bit_worker_process](https://www.terraform.io/docs/providers/azurerm/r/app_service.html#use_32_bit_worker_process)

jen20:
The use32BitWorkerProcess property does not resolve correctly using the default name mangling rules (producing "use32_bit_worker_process" instead of adding a _ between "use" and "32"), and changing the rules is likely a source of latent bugs (see discussion on pulumi/pulumi-terraform#350).

Instead, we map this field explicitly.

Fixes #157.

With this fix applied, the following program runs correctly:

```typescript
import * as azure from "@pulumi/azure";

// Create an Azure Resource Group
const resourceGroup = new azure.core.ResourceGroup("resourceGroup", {
    location: "West Europe",
});

const appServicePlan = new azure.appservice.Plan("test", {
    location: resourceGroup.location,

    sku: {
        tier: "Standard",
        size: "S1"
    }
});

const appService = new azure.appservice.AppService("test", {
    location: resourceGroup.location,
    appServicePlanId: appServicePlan.id,

    siteConfig: {
        dotnetFrameworkVersion: "v4.0",
        use32BitWorkerProcess: true,
        scmType: "LocalGit",
    },

    appSettings: {
        "SOMEKEY": "SomeValue",
    }
});
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,8 +1,8 @@
 ## 0.17.2 (Unreleased)
 
-### Improvements
-
 - Updated to v1.23.0 of the AzureRM Terraform Provider.
+
+- Fixed a bug where AppService or FunctionApp could not be created with `use32BitWorkerProcess` set to `true`.
 
 ## 0.17.1 (Released March 6, 2019)
 
--- a/resources.go
+++ b/resources.go
@@ -235,6 +235,15 @@ func Provider() tfbridge.ProviderInfo {
 					// Max length of a functionapp name is 60.
 					// This was discovered directly through the portal.
 					azureName: AutoNameWithMaxLength(azureName, 60),
+					"site_config": {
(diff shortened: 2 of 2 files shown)


jen20:
Build failures here are unrelated to the changes, I've restarted it.

PR #211, opened 2019-04-17
Addition of new Azure resources as added to terraform v1.25.0

stack72:
@jen20 this is what we were debugging together on zoom


resources.go
```diff
@@ -333,6 +337,16 @@ func Provider() tfbridge.ProviderInfo {
 			// DataBricks
 			"azurerm_databricks_workspace": {Tok: azureResource(azureDataBricks, "Workspace")},
 
+			// DataFactory
+			"azurerm_data_factory":                           {Tok: azureResource(azureDataFactory, "Factory")},
```
jen20: I wonder if we'd be better mapping this one as `Factory`, but removing that prefix from all of the other resources?

stack72: changed as requested!

jen20: This looks good to merge, pending CI.

Issue #213, opened 2019-04-25
Move serverless function support from pulumi-azure-serverless into pulumi-azure


lukehoban:
Like we did for AWS and GCP, we should move the core `CallbackFunction` concept and event handlers into the core `@pulumi/azure` package for consistency.  We should ideally move everything in `pulumi-azure-serverless` down and then deprecate it.  If that is not possible, we should likely renaming it to `azurex` as the initial foundation of a larger `azurex` package we'll build up in the near future again for consistency across our libraries.

PR #214, opened 2019-04-25
Moving azure-serverless directly down into pulumi/azure.

CyrusNajmabadi:
Fixes https://github.com/pulumi/pulumi-azure/issues/213

Note: this is not a direct port down.  I'm using this opportunity to:
1. clean thigns up.
2. remove deprecated functionality.
3. align more closely in naming/style with the corresponding entities in `@pulumi/aws`.


CyrusNajmabadi:
Note: this is **NOT** ready for review.  If ou want to skim leave comments, that's fine :)

CyrusNajmabadi:
Tagging @mikhailshilkov

examples/http/index.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+import * as eventhub from "@pulumi/azure/eventhub";
```
aydrian: Why are you importing `eventhub`? Maybe I'm not seeing it.

ellismg:
I think we should default to using V2 of the Functions runtime and also set the default version to either Node 8 or Node 10.  By default you get the V1 runtime and node 6.X, which is EOL in a few days.

ellismg:
I also wonder if we should adopt https://github.com/Azure/azure-functions-nodejs-worker/tree/dev/types/public instead of the current package we are using, as these seem to be the "official" type definitions.

mikhailshilkov:
V2 would require managing extensions for any non-HTTP functions, which means installing NuGet packages, including node functions. They are supposed to be working on removing this cumbersome dependency, so maybe worth waiting.

examples/http/index.ts
CyrusNajmabadi: Fixed!

CyrusNajmabadi:
>I also wonder if we should adopt https://github.com/Azure/azure-functions-nodejs-worker/tree/dev/types/public instead of the current package we are using, as these seem to be the "official" type definitions.

Yes!  thank you for finding this! :)

CyrusNajmabadi:
>V2 would require managing extensions for any non-HTTP functions, which means installing NuGet packages, including node functions.

Could you clarify this a bit more and/or including links to docs on this?

Note that we currently do install nuget packages for our AWS provider, and that's not really a problem for us.  Not sure what is meant by: "including node functions."

mikhailshilkov:
@CyrusNajmabadi "including node functions" means that binding extension packages are required to be in the deployment zip for Node.JS functions.

The link:
https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-register

But it looks like extension bundles are already in the docs, so maybe it's even easier than I thought.

CyrusNajmabadi:
Merging in.  Let me know if there are any other changes you want made.

PR #223, opened 2019-04-30
Configure URL and HTTP methods of HttpEventSubscription

### LINKED ISSUE #220: Configure URL and HTTP methods of HttpEventSubscription
1. Now, HTTP function always gets the name of `HttpEventSubscription` which isn't very flexible. E.g. I can't have two Function Apps with functions with same names because I can't name Function Apps the same. I may have multiple Apps each having a Function called "http", which isn't possible now.

2. We should use `route` parameter too.

Here is an example of HTTP binding which is not possible with the current serverless extensions:

``` json
{
      "type": "httpTrigger",
      "name": "req",
      "direction": "in",
      "authLevel": "anonymous",
      "route": "{id}",
      "methods": [
        "get"
      ]
    }
```

Note the route which basically allows to accept a parameter in the URL itself and get the parameter value in the binding data: `context.bindingData.id`.

3. While we are here, `methods` should also be configurable to allow POST requests.

mikhailshilkov:
Fixes #220 with explicit configuration parameters

### DIFF
--- a/sdk/nodejs/appservice/zMixins_http.ts
+++ b/sdk/nodejs/appservice/zMixins_http.ts
@@ -51,10 +51,24 @@ export type HttpEventSubscriptionArgs = util.Overwrite<mod.CallbackFunctionAppAr
      * used.
      */
     location?: pulumi.Input<string>;
+
+    /**
+     * Defines the route template, controlling to which request URLs your function responds. The 
+     * default value if none is provided is <functionname>.
+     */
+    route?: string;
+
+    /**
+     * An array of the HTTP methods to which the function responds. If not specified, the function 
+     * responds to all HTTP methods.
+     */
+    methods?: string[];
 }>;
 
 interface HttpBindingDefinition extends mod.BindingDefinition {
     authLevel?: "anonymous";
+    route?: string;
+    methods?: string[];
 }
 
 /**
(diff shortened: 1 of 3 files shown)


sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -51,10 +51,24 @@ export type HttpEventSubscriptionArgs = util.Overwrite<mod.CallbackFunctionAppAr
...
+     * Defines the route template, controlling to which request URLs your function responds. The 
+     * default value if none is provided is <functionname>.
+     */
+    route?: string;
+
+    /**
+     * An array of the HTTP methods to which the function responds. If not specified, the function 
+     * responds to all HTTP methods.
+     */
+    methods?: string[];
```
CyrusNajmabadi: these should be Inputs.  `methods` should be `Input<Input<string>[]>`.

CyrusNajmabadi:
Thanks! just needs a small tweak.

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: Do I understand correctly that then `const bindings` become `Input`, the parent constructor has to change, and so on throughout the chain?

PR #227, opened 2019-05-01
Simplify http code

CyrusNajmabadi:


CyrusNajmabadi:
Tagging @mikhailshilkov  :)

CyrusNajmabadi:
>Somehow I thought we don't want to mix Input types into the binding definition which partially comes from external. Looks simpler though, so LGTM.

Nope!  that's totally fine :)

PR #229, opened 2019-05-06
Allow configuring extension-specific host settings

mikhailshilkov:
I needed to configure `routePrefix` parameter of an HTTP function to get rid of the default `api` prefix and just go to the root.

So, I made an attempt to add [HTTP extension properties](https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json#http) to `host.json` file. If this change is good, I'll add settings to all other extensions that we have to far.


mikhailshilkov:
tagging @CyrusNajmabadi

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -104,11 +135,18 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
+                ...args.hostSettings,
+                extensions: { http: args.extensionSettings },
+            },
         }, opts);
 
-        this.url = args.route
-            ? pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/{${args.route}}`
-            : pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/${name}`;
+        const routePrefix = args.extensionSettings && args.extensionSettings.routePrefix;
+        const rootPath = routePrefix === '' ? '' : `${routePrefix || 'api'}/`;
```
CyrusNajmabadi: can you use "double quotes"?  also, it woudlbe preferred to write this as `routePrefix === undefined ? "api" : routePrefix`.

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -104,11 +135,18 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
             ...args,
             location,
             resourceGroupName,
+            hostSettings: {
+                ...args.hostSettings,
+                extensions: { http: args.extensionSettings },
```
CyrusNajmabadi: i'm a little confused by this. why can't hte caller just provide this 'extensionSetting' in the 'hostSettings object tehy can provide (since we just pass that along)

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -104,11 +135,18 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
+            },
         }, opts);
 
-        this.url = args.route
-            ? pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/{${args.route}}`
-            : pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/${name}`;
+        const routePrefix = args.extensionSettings && args.extensionSettings.routePrefix;
+        const rootPath = routePrefix === '' ? '' : `${routePrefix || 'api'}/`;
+
+        const functionPath = args.route ? `{${args.route}}` : name;
```
CyrusNajmabadi: i would explicitly check for 'undefined', not 'falsy'.

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: I started with that, but then I thought that passing Blob options doesn't make sense for HTTP trigger, so why expose the full surface of options when only a small subset is meaningful in this context.

CyrusNajmabadi: we don't have to expose the full set.  But we can say, for http events, that the args has a `hostSettings` object with the appropriate field.  One field would be `extensions` which has a `http` arg. 

Basically, from the TS perspective, we only expose what we think is relevant, but we keep the shape in line with the actual shape that is passed down.

mikhailshilkov: `hostSettings` is defined in `CallbackFunctionAppArgs`. Does it mean that `hostSettings` should become generic with a different type of `extension` in different event subscriptions?

CyrusNajmabadi: It doesn't even need to be generic (though that's one way to do this).  TS is *very* lenient here.  So just do this:

```ts
interface HttpHostSettings extends HostSettings {
   // add your new fields.
}

//...

export type HttpEventSubscriptionArgs = ... {
    hostSettings?: HttpHostSettings;
}>;
```

Basically, you're saying that for an http event subscription, the hostSettings property has these additional optional props.  This is extremely common in TypeScript and would be very much prefered here for me.

Does that make sense? :)  Sorry for the major delays in getting back to you.  I'm traveling and am on vacation :)

mikhailshilkov: Great! I still need to learn the TypeScript way of doing things. Thanks for explaining. Have a look at the latest version.
Enjoy your vacation :)

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -32,6 +32,34 @@ export type HttpRequest = azurefunctions.HttpRequest;
...
 export type HttpResponse = azureessentials.HttpResponse;
 
+/**
+ * Host settings specific to the HTTP plugin.
+ *
+ * For more details see https://docs.microsoft.com/en-us/azure/azure-functions/functions-host-json#http
+ */
+export interface HttpHostSettings extends mod.HostSettings {
+    extensions?: {
+        http: {
```
CyrusNajmabadi: would it make sense to have this be optional too?  or are we really saying "if you provide 'extensions' is *must* have an 'http' property in it"?

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -106,9 +140,12 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
 
-        this.url = args.route
-            ? pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/{${args.route}}`
-            : pulumi.interpolate`https://${this.functionApp.defaultHostname}/api/${name}`;
+        const routePrefix = args.hostSettings && args.hostSettings.extensions && args.hostSettings.extensions.http.routePrefix;
+        const rootPath = routePrefix === "" ? "" : `${routePrefix === undefined ? "api" : routePrefix}/`;
+
+        const functionPath = args.route === undefined ? name : `{${args.route}}`;
+
+        this.url = pulumi.interpolate`https://${this.functionApp.defaultHostname}/${rootPath}${functionPath}`;
```
CyrusNajmabadi: dumb question.  instead of us computing this ourselves, is it possible to obtain off of the FunctionApp in some way?  If not, this is fine.  but it would be great if we could just defer to Azure here instead of computing it ourselves.

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: For now, it's going to be the only property, so makes sense to require it IMO. Making it optional later shouldn't break anything, right?

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: I don't think we can get it directly. We'd need to do [something like this](https://stackoverflow.com/a/42940281/1171619).

sdk/nodejs/appservice/zMixins_http.ts
CyrusNajmabadi: it technically can.  but it's highly unlikely.  so i'm ok with it being required.

mikhailshilkov:
Added settings for queues and topics. Blobs don't seem to have any.

sdk/nodejs/storage/zMixins.ts
```diff
@@ -261,6 +261,32 @@ export interface QueueContext extends appservice.Context<void> {
     };
 }
 
+/**
+ * Host settings specific to the Storage Queue plugin.
+ *
+ * For more details see https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue#host-json
+ */
+export interface QueuesHostSettings extends appservice.HostSettings {
```
CyrusNajmabadi: Why 'Queues' instead of 'Queue'?

mikhailshilkov: Because the extension is called `queues`... But then I broke this rule with Topic naming. I'll rename to 'Queue'.

sdk/nodejs/appservice/zMixins_http.ts
CyrusNajmabadi: 👍

sdk/nodejs/storage/zMixins.ts
```diff
@@ -285,6 +311,12 @@ export type QueueEventSubscriptionArgs = util.Overwrite<appservice.CallbackFunct
      * used.
      */
     location?: pulumi.Input<string>;
+
+    /** 
+     * Host settings specific to the Storage Queue plugin. These values can be provided here, or defaults will 
+     * be used in their place. 
+     */
+    hostSettings?: QueueHostSettings;
 }>;
```
CyrusNajmabadi: can we extract teh topic/queue stuff out into separate PRs?

mikhailshilkov: We can... why?

CyrusNajmabadi: i just preferred keeping the commits nice and distinct :)  it's not a huge deal though!

mikhailshilkov: This is cohesive enough IMO. I'll do so next time!

mikhailshilkov:
@CyrusNajmabadi @lukehoban Anything preventing the merge of this PR?

CyrusNajmabadi:
Thanks!

Issue #218, opened 2019-04-26
Add support for more Azure FunctionApp triggers.


CyrusNajmabadi:
We currently support http/topic/container/queue triggers.  However, now that we're on azure functions 2.0 we can support the following as well:

1. [EventGrid triggers](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid#javascript-example)
2. [EventHub triggers](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-hubs#trigger---javascript-example)
3. [Timer triggers](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-timer)

mikhailshilkov:
+ 4. [Cosmos DB trigger](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-cosmosdb-v2)

mikhailshilkov:
So, the Event Grid trigger implementation is a bit more tricky:

1. In order to link an Event Grid subscription to an Azure Function, the webhook URL [must include a system key](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid#version-2x-runtime).

2. There is an HTTP API to get a system key, but it requires [a master key](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-http-webhook#authorization-keys).

3. None of the keys are available from Pulumi resources.

4. None of the keys are available from Terraform resources: [issue](https://github.com/terraform-providers/terraform-provider-azurerm/issues/699).

5. It's not there because Azure Go SDK [doesn't support it](https://github.com/Azure/azure-sdk-for-go/issues/2397).

6. [The last comment there](https://github.com/Azure/azure-sdk-for-go/issues/2397#issuecomment-477271267) mentions that there will be a new API available soon from the Functions side.

There is an alternative path: create an HTTP-triggered function and [handle EventGrid webhook](https://github.com/Azure/azure-sdk-for-go/issues/2397#issuecomment-477271267) in manual mode: respond to validation and parse the event from HTTP request.

The downside is that we get an open HTTP function hanging around. The upside: flexibility to get events in Cloud Events standard shape.

mikhailshilkov:
I'm watching https://github.com/Azure/azure-functions-host/issues/3994 which should simplify the implementation of the Event Grid trigger (no need to call Kudu APIs)

mikhailshilkov:
Everything here except EventGrid is done. Closing in favor of #295.

PR #236, opened 2019-05-09
Timer triggered functions

mikhailshilkov:
Implements part of #218


examples/timer/index.ts
```diff
@@ -0,0 +1,11 @@
+import * as azure from '@pulumi/azure';
+
+const resourceGroup = new azure.core.ResourceGroup('example', { location: 'West US 2' });
+
+new azure.appservice.TimerSubscription('everyminute', {
+    resourceGroup,
+    schedule: '0 * * * * *',
```
CyrusNajmabadi: in awsx we also allowed a richer data structure for schedules: https://github.com/pulumi/pulumi-awsx/blob/master/nodejs/awsx/autoscaling/schedule.ts

Can we also do something like that here?

examples/timer/index.ts
```diff
@@ -0,0 +1,11 @@
+import * as azure from '@pulumi/azure';
+
+const resourceGroup = new azure.core.ResourceGroup('example', { location: 'West US 2' });
+
+new azure.appservice.TimerSubscription('everyminute', {
+    resourceGroup,
+    schedule: { second: 0 },
```
CyrusNajmabadi: i'm not certain how `second: 0` translates to `everyminute`.

sdk/nodejs/appservice/zMixins_timer.ts
```diff
@@ -0,0 +1,249 @@
...
+    function checkRange(
+            val: number | undefined, name: keyof ScheduleArgs,
+            minInclusive: number, maxInclusive: number) {
+        if (val !== undefined) {
+            if (val < minInclusive || val > maxInclusive) {
+                throw new Error(`Value for [args.${name}] was not in the inclusive range [${minInclusive}, ${maxInclusive}].`);
+            }
+        }
+    }
+}
```
CyrusNajmabadi: pity we can't share all this code.

examples/timer/index.ts
mikhailshilkov: It translated to "0 * * * * *" which means "every time second = 0" means every minute. Not super obvious, like most cron expressions...

CyrusNajmabadi: for the purposes of an example how about `{ minute: 5 }` and call it `everyfiveminutes` :)

mikhailshilkov: `{ minute: 5}` will trigger when minute is 5, every second. So, 60 times per hour. Every five minutes is `0 */5 * * * *`.

CyrusNajmabadi: ok.  can we do `{ second: 0, minute: 5 }` :D

mikhailshilkov: That is going to run once per hour, at X:05:00

mikhailshilkov:
@CyrusNajmabadi @lukehoban Anything preventing the merge of this PR?

CyrusNajmabadi:
Had to fix a merge conflict.  Once this passes, i'll merge in.  If i forget def ping me!

PR #237, opened 2019-05-09
Cosmos triggered Functions

mikhailshilkov:
Part of #218


sdk/nodejs/cosmosdb/zMixins.ts
```diff
@@ -0,0 +1,203 @@
...
+
+    /**
+     * When set, it adds a prefix to the leases created in the Lease collection for this Function, effectively allowing 
+     * two separate Azure Functions to share the same Lease collection by using different prefixes.
+     */
+    leaseCollectionPrefix: pulumi.Input<string>;
+
+    /**
+     * When set to true, the leases collection is automatically created when it doesn't already exist. 
+     * The default value is false.
```
CyrusNajmabadi: it says default is false, but passes 'true' below.

sdk/nodejs/cosmosdb/zMixins.ts
```diff
@@ -0,0 +1,203 @@
...
+        const { resourceGroupName, location } = appservice.getResourceGroupNameAndLocation(args, account.resourceGroupName);
+
+        const bindingConnectionKey = "BindingConnectionAppSettingsKey";
+
+        const bindings: CosmosBindingDefinition[] = [{
+            name: "items",
+            direction: "in",
+            type: "cosmosDBTrigger",
+            leaseCollectionPrefix: name,
+            createLeaseCollectionIfNotExists: true,
```
CyrusNajmabadi: should this be user configurable?

sdk/nodejs/cosmosdb/zMixins.ts
mikhailshilkov: If we don't specify it in binding properties, Azure will assume `false`. That's why we pass `true` below. Too confusing?

sdk/nodejs/cosmosdb/zMixins.ts
mikhailshilkov: Do we want to expose all properties as configurable? (there are more of them) Or take some defaults for now, which make sense in 95% of cases? (my call)

sdk/nodejs/cosmosdb/zMixins.ts
CyrusNajmabadi: gotcha.  it is a bit confusing.  so we should just explain what's going on in the code (either here or at the poitn you pass in 'true').  we should also state why it's appropriate to pass in 'true' .

sdk/nodejs/cosmosdb/zMixins.ts
CyrusNajmabadi: we can totally not expose if we think this is reasonable.  users can always request in the future.  though we should likely doc why we think this is the right default.

mikhailshilkov: Added a comment to cover both.

henriktao:
Is onChange() still the recommended way of triggering functions through cosmos DB? I have encountered a problem that the lease collection is not created even though createLeaseCollectionIfNotExists is set to true.

PR #249, opened 2019-05-14
Fix the build by fixing naming in cosmos example

mikhailshilkov:
Looks I copy-pasted it wrong to the example and haven't run the final version.
Now it's detected by the test suite.
Fixing the sample.


mikhailshilkov:
Hasn't been able to figure out the latest issue with the build yet... Shows lots of changed files in the tree... Hints are welcome.

PR #255, opened 2019-05-21
Multi function app

### LINKED ISSUE #225: Multiple functions in a serverless app
With the latest azure serverless wrappers, each function is always deployed to a separate app.

Function Apps support multiple functions per app, and I think we should still allow this.

Multiple functions share the same resources, which means potentially lower cost and shared cold starts. People also do things like adding a dummy timer trigger to all HTTP functions to prevent some of the cold starts.

mikhailshilkov:
Fixes #225

A new component `MultiFunctionApp` (needs a better name) is introduced. It enables writing applications like this:

``` ts
const http1 = new azure.appservice.HttpFunction("hello", {
    callback: async (context, args) => {
        return {
            status: 200,
            body: "Hello World!"
        };
    }
});

const http2 = new azure.appservice.HttpFunction("rock", {
    callback: async (context, args) => {
        return {
            status: 200,
            body: "You ROCK!"
        };
    },
});

const timer1 = new azure.appservice.TimerFunction("warmer", {
    schedule: "0 */5 * * * *",
    callback: async () => {},
});
    
const app = new azure.appservice.MultiFunctionApp("myapp", {
    functions: [http1, http2, timer1],
});
```

This will result in a single Function App with three Azure Functions on it.

The old single-function API still works but delegates work to the multi-function component.

I've excluded output bindings from this PR, will take it separately.

Questions for the discussion:

1. Is this the right direction? Any design suggestions?
2. Any better way to unify single-function and multi-function models?
3. A function can be added to one Function App (most typical), but now it can be added to zero (by mistake) or N (e.g. warmer timer). Is there a good way to prevent zero usage?

Tagging @lukehoban @CyrusNajmabadi


sdk/nodejs/appservice/zMixins.ts
```diff
@@ -501,3 +501,264 @@ export function getResourceGroupNameAndLocation(
     const getResult = resourceGroupName.apply(n => core.getResourceGroup({ name: n }));
     return { resourceGroupName, location: getResult.location };
 }
+
+export type CallbackArgs<C extends Context<R>, E, R extends Result> = {
```
CyrusNajmabadi: Can we instead extract this part out from the existing CallbackFunctionAppArgs?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -501,3 +501,264 @@ export function getResourceGroupNameAndLocation(
...
+
+    if (!args.callback && !args.callbackFactory) {
+        throw new Error("One of [callback] or [callbackFactory] must be provided.");
+    }
+
+    const func = args.callback
+        ? redirectConsoleOutput(args.callback)
+        : () => redirectConsoleOutput(args.callbackFactory!());
+
+    return pulumi.output(pulumi.runtime.serializeFunction(func, { isFactoryFunction: !!args.callbackFactory }));
```
CyrusNajmabadi: similarly, we should use this as well instead of having a copy of this code elsewhere.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -501,3 +501,264 @@ export function getResourceGroupNameAndLocation(
...
+     * This form is useful when there is expensive initialization work that should only be executed
+     * once.  The factory-function will be invoked once when the final Azure FunctionApp module is
+     * loaded. It can run whatever code it needs, and will end by returning the actual function that
+     * the Azure will call into each time the FunctionApp it is is invoked.
+     */
+    callbackFactory?: CallbackFactory<C, E, R>;
+}
+
+export function serializeFunctionCallback<C extends Context<R>, E, R extends Result>(
+    args: CallbackArgs<C, E, R>): pulumi.Output<pulumi.runtime.SerializedFunction> {
```
CyrusNajmabadi: 1. consider juts mkaing this return a promise.  it doesn't really need to be an Output.
2. does tihs need to be exported?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -501,3 +501,264 @@ export function getResourceGroupNameAndLocation(
...
+            map[`${func.name}/handler.js`] = new pulumi.asset.StringAsset(func.func.text);
+        }
+
+        return map;
+    });
+}
+
+function combineAppSettings(args: MultiFunctionAppArgs): pulumi.Output<{[key: string]: string}> {
+    
+    const perFunctionSettings = args.functions !== undefined ? args.functions.map(c => c.definition.apply(b => b.appSettings)) : [];
```
CyrusNajmabadi: would it ever make sense to have a MultiFunctionApp without 'args.functions'?  if not, we should just make it non-optional.

mikhailshilkov: An alternative is to provide an Archive of a folder on the disk (e.g. for .NET/Java functions), and yet avoid setting Plan/Storage/Blob/settings manually.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: `CallbackArgs` is intended for an individual Azure Function, so all other parameters of `FunctionAppArgs` do not apply. I agree this deserves a clean-up, eventually.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -501,3 +501,264 @@ export function getResourceGroupNameAndLocation(
...
+    return pulumi.all([args.appSettings, ...perFunctionSettings]).apply(items => items.reduce((a, b) => ({ ...a, ...b }), {}));
+}
+
+export type MultiFunctionAppArgs = util.Overwrite<FunctionAppArgs, {
+
+    /**
+     * The functions to deploy as parts of this application.
+     */
+    functions?: AzureFunction[];
```
CyrusNajmabadi: should this be optional?

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: What i'm saying is:  Since you extracted this, teh normal CallbackFunctionAppArgs should then derive from this type, so we don't have the duplication.  Thanks! :)

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Changed to Promise and marked as internal (needs to be accessed from other mixin files)

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: How do I derive `CallbackFunctionAppArgs` from FunctionAppArgs and CallbackArgs with util.Overwrite? Do I need an intermediate
```
interface FunctionAppArgsWithCallback<C extends Context<R>, E, R extends Result> 
    extends FunctionAppArgs, CallbackArgs<C, E, R> {
}
export type CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> = 
    util.Overwrite<FunctionAppArgsWithCallback<C, E, R>, {
...
```
?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -246,81 +255,257 @@ export interface HostSettings {
...
+    constructor(type: string, name: string,
+                func: FunctionArgs,
+                args: CallbackFunctionAppArgs<C, E, R>,
+                opts: pulumi.ComponentResourceOptions = {}) {
+        super(type, name, undefined, opts);
+
+        const multiArgs = {
+            ...args,
+            functions: [func],
+        };
```
CyrusNajmabadi: consider just inlining this.

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: So, i'm thinkin like this (names not final):

```ts
interface BaseFunctionAppArgs<C, E, R> {
    callback?: Callback<C, E, R>;
}

type CallbackFunctionAppArgs<C, E, R> = BaseFunctionAppArgs<C, E, R> & utils.Overwrite<...
```

Note: we should likely move away from utils.Overwrite for this package as well (we already did for awsx).  THe problem is that it makes the docs and type-sigs really hard for normal mortals to read :)   We isntead use interfaces.  This means we do need to copy over the values we care about from teh core interfaces, but it makes it super easy to grok.  In this case, it would then look like:

```ts
interface BaseFunctionAppArgs<C, E, R> {
    callback?: Callback<C, E, R>;
}

interface CallbackFunctionAppArgs<C, E, R> extends BaseFunctionAppArgs<C, E, R>, SomeOtherCommonInterface {
     // add additional members here manually
}
```

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -246,81 +255,257 @@ export interface HostSettings {
...
+export function serializeFunctionCallback<C extends Context<R>, E, R extends Result>(
+    args: CallbackArgs<C, E, R>): Promise<pulumi.runtime.SerializedFunction> {
 
     if (args.callback && args.callbackFactory) {
         throw new pulumi.RunError("Cannot provide both [callback] and [callbackFactory]");
     }
 
     if (!args.callback && !args.callbackFactory) {
-        throw new Error("One of [callback] or [callbackFactory] must be provided.");
+        throw new pulumi.RunError("One of [callback] or [callbackFactory] must be provided.");
```
CyrusNajmabadi: Note: run-error is wrong in both cases.  It should just be error.  RunError is basically when we do *not* want the stack, because we can just give a reasonable message that needs no additional context (like: you need to provide config value X).  For arg validation like this, we actually want a normal error so you get a good stack making it clear what codepath is causing you to hit this.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -246,81 +255,257 @@ export interface HostSettings {
...
-function redirectConsoleOutput<C extends Context<R>, E, R extends Result>(callback: Callback<C, E, R>) {
-    return (context: C, event: E) => {
-        // Redirect console logging to context logging.
-        console.log = context.log;
-        console.error = context.log.error;
-        console.warn = context.log.warn;
-        // tslint:disable-next-line:no-console
-        console.info = context.log.info;
+async function produceDeploymentPackage(args: MultiFunctionAppArgs): Promise<pulumi.asset.AssetMap> {
+
```
CyrusNajmabadi: extra newline.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -246,81 +255,257 @@ export interface HostSettings {
...
 
-        return map;
-    });
+    /**
+     * HTTP route to call this function (applies to functions reachable via HTTP).
+     */
+    route?: pulumi.Input<string>;
 }
 
-function redirectConsoleOutput<C extends Context<R>, E, R extends Result>(callback: Callback<C, E, R>) {
```
CyrusNajmabadi: note: consider restoring the positoin of some of these helpers to make the diff easier :)

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -246,81 +255,257 @@ export interface HostSettings {
...
+
+    /**
+     * The functions to deploy as parts of this application.
+     */
+    functions?: FunctionArgs[];
+
+    /**
+     * The deployment package to deploy as-is (instead of the 'functions' property).
+     */
+    archive?: pulumi.Input<pulumi.asset.Archive>;
```
CyrusNajmabadi: consider explicitly doc'ing the relationship between props.  i.e.: Either one of [functions] or [archive] must be provided.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -334,139 +519,119 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
-    constructor(name: string, bindings: pulumi.Input<BindingDefinition[]>,
-                args: CallbackFunctionAppArgs<C, E, R>, opts: pulumi.CustomResourceOptions = {}) {
+        if (!args.functions && !args.archive) {
+            throw new pulumi.RunError("One of [functions] or [archive] must be provided.");
+        }
 
-        if (!args.resourceGroupName) {
-            throw new pulumi.ResourceError("[args.resourceGroupName] must be provided", opts.parent);
+        if (args.functions && args.archive) {
+            throw new pulumi.RunError("Cannot provide both [functions] and [archive]");
```
CyrusNajmabadi: both of these should be ResourceErrors.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -334,139 +519,119 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
         }
 
-        if (!args.location) {
-            throw new pulumi.ResourceError("[args.location] must be provided", opts.parent);
+        if (args.functions) {
+            const names = args.functions.map(f => f.name);
+            const duplicates = names.filter((item, index) => names.indexOf(item) !== index);
```
CyrusNajmabadi: clever!

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -334,139 +519,119 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
+        const rootPath = routePrefix === "" ? "" : `${routePrefix === undefined ? "api" : routePrefix}/`;
+
+        this.endpoints = this.app.defaultHostname.apply(hostName =>             
+            (args.functions !== undefined ? args.functions : [])
+                .filter(f => f.route !== undefined)
+                .map(f => [f.name, `https://${hostName}/${rootPath}${f.route}`])
+                .reduce((map, [key, val]) => {
+                    map[key] = val;
+                    return map;
+                }, {} as { [functioName: string]: string })
```
CyrusNajmabadi: this got a little too clever :)   can you just convert to a normal block? :)

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -37,27 +37,28 @@ export type HttpResponse = azureessentials.HttpResponse;
...
-            /**
-             * When enabled, this setting causes the request processing pipeline to periodically check system performance 
-             * counters like connections/threads/processes/memory/cpu/etc. and if any of those counters are over a built-in 
-             * high threshold (80%), requests will be rejected with a 429 "Too Busy" response until the counter(s) return 
-             * to normal levels.
-             */
-            dynamicThrottlesEnabled?: boolean,
-        }
-    }    
+        http: HttpHostExtensions
```
CyrusNajmabadi: nit.  have comma at end.

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -121,32 +122,73 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
-                    route: args.route,
-                    methods: args.methods,
-                }, {
-                    type: "http",
-                    direction: "out",
-                    name: "$return",
-                }];
-
-        super("azure:appservice:HttpEventSubscription", name, bindings, {
+        const func = new HttpFunction(name, args);
```
CyrusNajmabadi: is this exposed through teh base type?

sdk/nodejs/appservice/zMixins_timer.ts
```diff
@@ -228,22 +228,60 @@ export class TimerSubscription extends mod.EventSubscription<TimerContext, Timer
...
+    /**
+     * If true, the function is invoked when the runtime starts.
+     */
+    runOnStartup?: pulumi.Input<boolean>;
+}>;
+
+/**
+ * Azure Function triggered on a CRON schedule.
+ */
+export class TimerFunction implements mod.FunctionArgs {
```
CyrusNajmabadi: why do we need this actual class?

CyrusNajmabadi: ah, i see now.  so the idea is that a user can make a MultiFunctionApp by passing in these classes.  And these classes are sometimes thin shims, but sometimes contain a lot of logic.  👍  on this.  


Note: we should probably doc at teh MultiFunctionApp level how you're supposed to do this.

sdk/nodejs/storage/zMixins.ts
```diff
@@ -397,40 +440,71 @@ export class QueueEventSubscription extends appservice.EventSubscription<QueueCo
...
+     * Defines the queue to trigger the function.
+     */
+    queue: Queue;
+}>;
 
-        const bindings: QueueBindingDefinition[] = [{
+/**
+ * Azure Function triggered by a Storage Queue.
+ */
+export class QueueFunction implements appservice.FunctionArgs {
```
CyrusNajmabadi: given that this impls .FunctionArgs, we probably want to call this QueueFunctionArgs.  I think that helps explain what the purpose of this thing is.  It's the args that describe how to build a queue function, and those args will be given to a MultiFunctionApp so it can then actually build the real things.

CyrusNajmabadi:
@mikhailshilkov I very much like tihs.  Nice design.  I think it would be helpful if we could make this into two commits.  One is just the introduction of the MultiFunctionApp and the surrounding interfaces/types/classes that form the scaffolding for it.  The second would be to add the ability for all the surrounding areas to work with the MultiFunctionApp paradigm.  i.e. the changes to queues/timers etc.  Would you be ok breaking things apart like that?  I can try to help with this if you'd like!

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: `HttpFunction` is exposed for Multi-app

mikhailshilkov:
@CyrusNajmabadi Great, thanks for your comments! I'll break it down tomorrow.
I'm not a fan of `MultiFunctionApp` name... Any better ideas?
Ideally, it would be just `FunctionApp` but that's confusing because of the raw resource.

CyrusNajmabadi:
> @CyrusNajmabadi Great, thanks for your comments! I'll break it down tomorrow.
I'm not a fan of MultiFunctionApp name... Any better ideas?

Honestly, i don't mind.  I think it easily conveys what it's for and why you'd use it :)

sdk/nodejs/storage/zMixins.ts
mikhailshilkov: There's actually a thing called `QueueFunctionArgs`: it's the type of the second parameter of the constructor. I like calling them "XyzFunction" - read nice at the client side. Maybe, I'll rename `FunctionArgs` instead.

CyrusNajmabadi: I see.  that's reasonable.  We can say that a Function-App is built up out of Functions.  WFM :)

mikhailshilkov:
Closing in favour of #264

PR #259, opened 2019-05-24
Re-implement signedBlobReadUrl with a getAccountSAS call

### LINKED ISSUE #258: Expose signedBlobReadUrl function
Any reason why `azure.appservice.signedBlobReadUrl` is marked as internal?

We have its copy in [several](https://github.com/pulumi/examples/blob/master/azure-ts-appservice/sas.ts#L7) [similar](https://github.com/pulumi/examples/blob/master/azure-ts-functions/sas.ts#L7) [examples](https://github.com/pulumi/examples/blob/master/azure-ts-appservice-devops/infra/sas.ts#L7) and people are probably copying it all over the place.

mikhailshilkov:
Fixes #258


sdk/nodejs/storage/zMixins.ts
```diff
@@ -14,14 +14,64 @@
...
+                    delete: false,
+                    list: false,
+                    add: false,
+                    create: false,
+                    update: false,
+                    process: false,
+                }
+            });
+
+            return pulumi.output(accountSAS).apply(sas => `https://${accountName}.blob.core.windows.net/${containerName}/${blobName}${sas.sas}`);
```
ellismg: Could we instead just make the entire lambda we pass to the outer apply `async` and use `await` to resolve the promise returned by `getAccountSAS`, or just call `.then` on the promise, instead of doing this inner output and apply?

mikhailshilkov:
> Do you know if this changes the format of the code URL?

Yes, it will, the URL will be different. Both parameter order and values are different.

Note that we are using a SAS on a different level - account not a specific blob. Even container SAS [isn't supported](https://github.com/terraform-providers/terraform-provider-azurerm/issues/3068), and I don't even see an issue for blob-SAS. This puts us into a worse position security-wise. Should be fine for accounts auto-created inside our Serverless code, but doesn't feel great anyway.

sdk/nodejs/storage/zMixins.ts
```diff
@@ -14,14 +14,62 @@
...
+export function signedBlobReadUrl(blob: Blob | ZipBlob, account: Account): pulumi.Output<string> {
+
+    // Choose a fixed, far-future expiration date for signed blob URLs. The shared access signature
+    // (SAS) we generate for the Azure storage blob must remain valid for as long as the Function
+    // App is deployed, since new instances will download the code on startup. By using a fixed
+    // date, rather than (e.g.) "today plus ten years", the signing operation is idempotent.
+    const signatureExpiration = "2100-01-01";
+
+    return pulumi.all([account.name, account.primaryConnectionString, blob.storageContainerName, blob.name]).apply(
+        ([accountName, connectionString, containerName, blobName]) =>
```
CyrusNajmabadi: this could jus be async, and then you can await getAccountSAS below, right?

sdk/nodejs/storage/zMixins.ts
```diff
@@ -14,14 +14,62 @@
...
+        ([accountName, connectionString, containerName, blobName]) => 
+            getAccountSAS({
+                connectionString,
+                start: "2019-01-01",
+                expiry: signatureExpiration,
+                services: {
+                    blob: true,
+                    queue: false,
+                    table: false,
+                    file: false,
```
CyrusNajmabadi: is it required to be explicit on all the falses?

mikhailshilkov: Yes, they are all just `boolean`

sdk/nodejs/storage/zMixins.ts
mikhailshilkov: Yes... you don't like `then`?

CyrusNajmabadi: i'd prefer us use the more modern JS/TS primitives if we can :)  Is that ok with you here?

sdk/nodejs/storage/zMixins.ts
CyrusNajmabadi: WFM!

sdk/nodejs/storage/zMixins.ts
```diff
@@ -40,8 +40,8 @@ export function signedBlobReadUrl(blob: Blob | ZipBlob, account: Account): pulum
     const signatureExpiration = "2100-01-01";
 
     return pulumi.all([account.name, account.primaryConnectionString, blob.storageContainerName, blob.name]).apply(
-        ([accountName, connectionString, containerName, blobName]) => 
-            getAccountSAS({
+        async ([accountName, connectionString, containerName, blobName]) => {
```
CyrusNajmabadi: thanks!

CyrusNajmabadi:
Thanks much!

PR #264, opened 2019-05-29
Multi Function App

mikhailshilkov:
The first part of #255, the groundwork for #225:
- Introduces MultiFunctionApp which can hold several functions or accept an archive
- Redirects all existing callback-based functions to MultiFunctionApp with a single function in it
- Adds an example of an app with external code in PowerShell

The second part will contain all the specific Functions like http, timers, etc.


examples/http-external/funcapp/HelloPS/run.ps1
```diff
@@ -0,0 +1,28 @@
...
+else {
+    $status = [HttpStatusCode]::BadRequest
+    $body = "Please pass a name on the query string or in the request body."
+}
+
+# Associate values to output bindings by calling 'Push-OutputBinding'.
+Push-OutputBinding -Name Response -Value ([HttpResponseContext]@{
+    StatusCode = $status
+    Body = $body
+})
```
CyrusNajmabadi: cool!  great to have a PS example.  could we put this in hte examples repo later as well?

examples/http-external/index.ts
```diff
@@ -0,0 +1,15 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+
+const resourceGroup = new azure.core.ResourceGroup('example', { location: 'West US 2' });
```
CyrusNajmabadi: perhaps the location should come from config?  i don't care much though.

examples/http-external/index.ts
```diff
@@ -0,0 +1,15 @@
...
+// Create a Function App implemented in PowerShell with source code from 'funcapp' folder
+const app = new azure.appservice.MultiFunctionApp("http-ps", {
+    resourceGroupName: resourceGroup.name,
+    archive: new pulumi.asset.FileArchive("./funcapp"),
+    appSettings: {
+      "runtime": "powershell",
+    }
+});
+
+export const url = app.endpoints.apply(ep => `${ep[0]}HelloPS?name=Pulumi`);
```
CyrusNajmabadi: not quite clear to me why a MultiFunctionApp (with no additional args) would end up with an endpoint.  Is that normal behavior?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -88,7 +89,9 @@ export type CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> =
      * the Azure will call into each time the FunctionApp it is is invoked.
      */
     callbackFactory?: CallbackFactory<C, E, R>;
+}
 
+export type CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> = CallbackArgs<C, E, R> & util.Overwrite<FunctionAppArgs, {
```
CyrusNajmabadi: note: since we're making this cahnge, we probably should just exract out what we need from FunctionAppArgs and place directly in here.  it will fit the awsx approach.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -334,87 +444,123 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
+
+        this.endpoints = this.app.defaultHostname.apply(hostName => {
+            const endpoints = <{ [functionName: string]: string }>{};
+            if (args.functions !== undefined) {
+                const functionsWithEndpoints =  args.functions.filter(f => f.route !== undefined);
+                for (const f of functionsWithEndpoints) {
+                    endpoints[f.name] = `https://${hostName}/${rootPath}${f.route}`;
+                }
+            } else {
+                endpoints[0] = `https://${hostName}/${rootPath}`;
```
CyrusNajmabadi: this seems a bit wonky... would someone have to know to look this up?  should we instead have a defaultEndpoint for the case of an archive, and where the first func endpoint would go?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -423,15 +569,24 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
         super(type, name, undefined, opts);
 
-        this.functionApp = new CallbackFunctionApp(name, bindings, args, { parent: this });
+        const multiArgs = {
+            ...args,
+            functions: [{
+                name,
+                bindings,
+                body: serializeFunctionCallback(args),
+                route: name, // this is not entirely correct, will be fixed with a specific implementation of HttpFunction
```
CyrusNajmabadi: not certain i understand this comment.

CyrusNajmabadi:
Looking good.  need to go over this again in more detail later today.  had a couple of minor questions to that i'd like some clarity on.  but should be able to get this in in the next 24 hours.

examples/http-external/index.ts
mikhailshilkov: Every Function App has a default endpoint like `https://bla.azurewebsites.net/api/`. If no functions exist in the app, the endpoint will return 404. Functions can handle either `https://bla.azurewebsites.net/api/MyFunctionName` or any route if the route pattern is specified.

Even for an archive-based app, we should return an endpoint back for convenience.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Agreed, I'm not particularly happy about these endpoints...
Options:
1. Always return just 1 root endpoint and leave routing to users. A bit rudimentary: people might be used to seeing all URLs in Functions CLI / Visual Studio tools.
2. Somehow propagate URLs back to function objects? How? What if a function is in two apps?
3. Use a more sophisticated structure instead of key-value map.
4. ?

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: The route of HTTP functions depends on their configuration - you can check the other PR's `HttpFunction` class which passes this parameter down. I can't see how to implement this without making `HttpFunction` in this PR.

examples/http-external/funcapp/HelloPS/run.ps1
mikhailshilkov: yes, my todo list has an example of Functions in all supported languages

examples/http-external/index.ts
mikhailshilkov: I think it's hard-coded in all other samples in this repo

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: 1. i don't mind having the mapping object like you do.  but i would only do it for named key/value mappings.  I wouldn't have the `0` mapping.

You could also consider adding a `getEndpoint(data: Archive | Function)` which would the allow the user to pass in the same object they created the functionapp from to get the endpoint.  thougts?

mikhailshilkov: `getEndpoint` makes sense. Maybe just `getEndpoint(Function?)` - undefined giving you root?

CyrusNajmabadi: that sounds great!

CyrusNajmabadi:
@mikhailshilkov Let me know if you want another review!

mikhailshilkov:
@CyrusNajmabadi yes please!

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -334,87 +432,125 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
                 };
             }),
-        }, opts);
+        }, { parent: this });
 
         this.account = account;
         this.container = container;
         this.plan = plan;
         this.zipBlob = zipBlob;
+        this.functions = args.functions;
```
CyrusNajmabadi: should we expose hte optional archive for consistency here?

mikhailshilkov: I actually made functions private, thinking to use them in `getEndpoint`, but ended up not using them. I think it might be confusing to expose them and then not filling them in case of the archive scenario. Removed for now.

sdk/nodejs/appservice/zMixins_timer.ts
```diff
@@ -203,13 +203,6 @@ export type TimerSubscriptionArgs = util.Overwrite<mod.CallbackFunctionAppArgs<T
      */
     resourceGroupName?: pulumi.Input<string>;
 
-    /**
-     * Specifies the supported Azure location where the resource exists. Changing this forces a new
-     * resource to be created.  If not supplied, the location of the provided ResourceGroup will be
-     * used.
-     */
-    location?: pulumi.Input<string>;
```
CyrusNajmabadi: out of curiosity, why did all the 'location' stuff get removed?

mikhailshilkov: It used to be an override which would change the type to allow undefined. Meanwhile, location got to allow undefined in the base class, so overrides can be removed.

CyrusNajmabadi: got it, thanks!

mikhailshilkov:
@CyrusNajmabadi Time to merge yet?

lukehoban:
If I understand correctly - we're making the following changes:
1. Removing the previously pubic `CallbackFunctionApp<C,E,R>` subclass of `FunctionApp`
2. Adding a new  `MultiFunctionApp` component which builds a `FunctionApp` child
3. Changing `EventSubscription<C,E,R>`'s `functionApp` property to return a `MultiFunctionApp` instead of a `CallbackFunctionApp<C, E, R>`.

Are we okay with this level of breaking change?  If so, we should also add the breaking change label here and clarify the break in the `CHANGELOG`.

Was there a reason we needed to do (1) and (3)?  It's not clear to me how those were necessary to get the benefits here?

How exactly does `MultiFunctionApp` differ from `FunctionApp`?  Don't all Azure `FunctionApp`s already handle multiple functions?  If the distinction is not about multiple vs. single, then the naming is possible confusing since those are the two names that will be public classes for working with `FunctionApp`s.    (A comment on `MultiFunctionApp` would also help here).

I notice that `MultiFunctionApp`s allow passing in `body: Promise<pulumi.runtime.SerializedFunction>`.  I suspect we don't want to put that on public API surface area, and instead want to accept a JavaScript `Function` which we can internally serialize.  From the implementation, it looks like that should be possible.

mikhailshilkov:
@lukehoban Thanks a ton for your review and the great comments. I had most of those questions in my head but failed to answer them in a satisfactory way. Let's try again together.

> Are we okay with this level of breaking change?

Honestly, I did not prioritize avoiding breaking changes, thinking that this area of custom API for functions was under flux for the recent months. Do you think we might actually break some substantial number of users by this change? Or are breaking changes just wrong for the public product like Pulumi?

> Was there a reason we needed to do (1) and (3)?

I see two ways of not doing (1) and (3):

1. Adapt an existing `CallbackFunctionApp<C,E,R>` for the new needs. I can't see how that's possible because the new component should have no type parameters: the types are specific to a function and can be different in the same app.

2. Keep `CallbackFunctionApp` but add a new component side-by-side. I started with this in my first PR, but then we moved towards avoiding duplication.

> How exactly does MultiFunctionApp differ from FunctionApp

`FunctionApp` is a raw resource while `MultiFunctionApp` is a component which creates many sub-resources like an app, a storage account/container/blob, and a function app. Logically, they are the same thing, but on different levels. One of my ideas was to name it `FunctionApp` but in a separate module.

> Don't all Azure FunctionApps already handle multiple functions? 

They do. That's why we are busy extending API to accommodate this built-in feature.

> If the distinction is not about multiple vs. single, then the naming is possible confusing since those are the two names that will be public classes for working with FunctionApps. 

Agreed! I tried to express my lack of confidence about the name but failed to find a better one.

> I notice that MultiFunctionApps allow passing in body: Promise<pulumi.runtime.SerializedFunction>. I suspect we don't want to put that on public API surface area, and instead want to accept a JavaScript Function which we can internally serialize. 

Nice one! Other than `Function`, we seem to need `{ isFactoryFunction: !!args.callbackFactory }` for serialization. Is there a type which would hold this distinction? We also wrap functions with `redirectConsoleOutput` which expect `Callback<C,E,R>` and not a generic `Function`.

But agree we need a change, also to make `serializeFunctionCallback` private instead of internal - doesn't feel great atm.

Thanks again!

mikhailshilkov:
Maybe we shouldn't export `interface Function` at all, and the app would expect an array of functions each being a case in a discriminated union? I don't expect clients to supply their own types of Azure Functions, so we might just make this type closed. Is that somewhat common in the TypeScript world?

CyrusNajmabadi:
> Was there a reason we needed to do (1) and (3)? It's not clear to me how those were necessary to get the benefits here?

I clearly missed this in my review as well.  And i agree with Luke on this.  I think it would be better to just introduce MultiFunctionApp and keep around CallbackFunctionApp.  The latter would stay around for teh common case of just easily wanting to make a function-app out of a simple callback, and would likely be all that's necessary when hooking up a subscription for a particular resource.

I think MultiFunctionApp serves a great purpose, and conveys that it's primarily around to help have an easy way to combine several 'functions' into a single FunctionApp.

CyrusNajmabadi:
Toward your questions @mikhailshilkov 

> Do you think we might actually break some substantial number of users by this change? Or are breaking changes just wrong for the public product like Pulumi?

Ideally we avoid breaking changes.  There are a few types of breaks we need to be strongly concerned about:

1. source level breaks
2. binary level breaks.

IMO the latter is usually worse.  i.e. if we're not breaking the binary side, then people can still have their code work, even if it means that they'd need to tweak things in source when doing a recompile.  However, if it's not super importnt, we should try to avoid source breaks as well.  In this case, it looks like you'd have both breaks.  i.e. if someone directly instantiated a CallbackFunctionApp, that would now break as that class/constructor-function is just not there anymore.  

> Keep CallbackFunctionApp but add a new component side-by-side. I started with this in my first PR, but then we moved towards avoiding duplication.

I'd like this, but i think we can have the best of both worlds.  We can either extract out a common base that both CallbackFunctionApp and MultiFunctionApp use.  Or we can just extract out common code into helper functions that they both can call into.

CyrusNajmabadi:
> Agreed! I tried to express my lack of confidence about the name but failed to find a better one.

This is my bad.  Thinking about this, i realize it should just be called `FunctionApp`.  We have ample precedence for this in awsx.  It's common for our 'nice abstraction' type to be called the exact same thing, but hten wrap and expose the raw resource.  For example `awsx.ecs.Service` wraps the raw `aws.ecs.Service`.  I'm not sure why i thought it was necessarily to have a different name here, so apologies if i led you astray.

CyrusNajmabadi:
@mikhailshilkov Let me know if you need any more info or have additional questions.  I'm happy to help with this.

mikhailshilkov:
> it should just be called FunctionApp

But we don't have `awsx`-like module here. Now, `MultiFunctionApp` is in the same `azure.appservice` module. Where do we put this `FunctionApp`? Create `azure.functions.*`? But then all the existing stuff has to stay?

mikhailshilkov:
Another question - inherit from the raw resource, or inherit from `ComponentResource`? I felt the latter is more natural, but `CallbackFunctionApp` does the former...

mikhailshilkov:
@CyrusNajmabadi I tried to address the issues with the latest commit - all except the naming. Please have a look.

mikhailshilkov:
In my latest commits, I changed `MultiFunctionApp` to inherit from `FunctionApp` so that `CallbackFunctionApp` could inherit in its turn.

The downside is that now all helper resources (plan, account, etc.) are not children of `MultiFunctionApp`. `EventSubscription` component compensates this for `CallbackFunctionApp` but not for `MultiFunctionApp`.

stack72:
Hi @CyrusNajmabadi 	/ @lukehoban 	

Are you both happy with this approach? @mikhailshilkov 	if so, can you rebase this and we can get it merged?

Paul

CyrusNajmabadi:
Looking now!

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -312,15 +374,45 @@ function redirectConsoleOutput<C extends Context<R>, E, R extends Result>(callba
...
+     */
+    functions?: Function[];
+
+    /**
+     * The deployment package to deploy as-is. Either one of [functions] or [archive] must be provided.
+     */
+    archive?: pulumi.Input<pulumi.asset.Archive>;
+};
+
+export class MultiFunctionApp extends FunctionApp {
```
CyrusNajmabadi: ok... so here's where i'm landing with this.  

First: because `MultiFunctionApp` takes in an array of `Function`s, and because each `Function` wraps a `Callback`, I think this really is a `MultiCallbackFunctionApp`.  (Or it's just `CallbackFunctionApp`, and we make it so that `CallbackFunctionApp` can take one or many callbacks).

Second: II think 'archive' muddies the waters here and makes it less clear what hte purpose of this all is.  I think i'd be fine with ArchiveFunctionApp being something else entirely.  That way there are clear, separated purposes for all these types.  the `CallbackFunctionApp` (and *possibly* `MultiCallbackFunctionApp` if we have that) is for wrapping up 1-many callbacks into a FunctionApp.  `ArchiveFunctionApp` is for the archive case.



Now, i'd love if we just ended up with CallbackFunctionApp allowing 1-many functions (with constructor overloads for the existing constructor we support).  If that's not possible, i'm ok having that exist for the '1-case', and MultiCallbackFunctionApp exist for the '2-N case'.

Thoughts?

CyrusNajmabadi:
Sorry for the massive delays.  I think i have a formalization that i like that i've laid out above.  First, let me know if my position is clear.  If not, i'm happy to expand more.  Second, let me know if my position is sane :)

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: How do we unify a generic `CallbackFunctionApp<C, E, R>` with the non-generic `CallbackFunctionApp` with multiple functions?

CyrusNajmabadi:
Ok.  I also talked with Luke (who is super busy, so i'll try to not have to pull him in directly).  He thought you were trying to take an approach that would support 'multiple languages' in a single function app.  

I can see through the 'powershell' example that a different language is def supported through the 'archive' approach.  However, is this that something you want to provide in a form that's more flexible than just hte large "bulk archive"?   i.e. would people have an easier way of providing the independent different-language functions here?  

Thanks!  Also, happy to chat on slack if htat will help out things here.

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: @mikhailshilkov Good point.  We likely can't.  :)  Based on that, it seems ok to have the split there.

CyrusNajmabadi: Note: i think splitting out 'archive' would be fine.

mikhailshilkov:
@CyrusNajmabadi I've made the change that we discussed yesterday. 

It happened to be a bit more complicated. I have two hierarchies of classes:

1. Derived from the `FunctionApp` resource: a helper private `CombinedFunctionApp` and `CallbackFunctionApp` that derives from it. That's to keep `CallbackFunctionApp` intact but reuse some of the code for (2).

2. Derived from `ComponentResource`: `ArchiveFunctionAppBase` base component and its children `ArchiveFunctionApp` and `MultiCallbackFunctionApp`. These are needed to contain all the resources under a single parent so that the resulting Pulumi tree is nice a clean.

Let me know if you see a way to simplify or any other remarks.

CHANGELOG.md
```diff
@@ -43,6 +46,7 @@ CHANGELOG
   4. You can subscribe to `azure.storage.Container` blob events, invoking a FunctionApp containing the code you want, using `container.onBlobEvent(...)`.
   5. You can subscribe to `azure.storage.Queue` events, invoking a FunctionApp containin the code you want, using `queue.onEvent(...)`.
 * The location for each resource attached to a provider instance is now set as part of the provider configuration. This means that the following code from v0.18.0:
+>>>>>>> upstream/master
```
CyrusNajmabadi: oops :)

examples/http-external/index.ts
```diff
@@ -0,0 +1,15 @@
+import * as pulumi from "@pulumi/pulumi";
+import * as azure from "@pulumi/azure";
+
+const resourceGroup = new azure.core.ResourceGroup('example', { location: 'West US 2' });
+
+// Create a Function App implemented in PowerShell with source code from 'funcapp' folder
+const app = new azure.appservice.ArchiveFunctionApp("http-ps", {
```
CyrusNajmabadi: cool.  i think this reads well.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -312,15 +370,56 @@ function redirectConsoleOutput<C extends Context<R>, E, R extends Result>(callba
...
+     * The deployment package of a Function App to deploy as-is.
+     */
+    archive: pulumi.Input<pulumi.asset.Archive>;
+};
+
+/**
+ * A composite resouce which creates a Function App and all required resources that the app depends on.
  */
-export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> extends FunctionApp {
+class CombinedFunctionApp extends FunctionApp {
```
CyrusNajmabadi: should this be abstract?  we woudln't expect this to be instnatiated directly right?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -415,6 +515,128 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
+ */
+export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> extends CombinedFunctionApp {
+    constructor(name: string, bindings: pulumi.Input<BindingDefinition[]>,
+                args: CallbackFunctionAppArgs<C, E, R>, opts: pulumi.CustomResourceOptions = {}) {
+        const functionsArgs = {
+            ...args,
+            functions: [<Function>{
+                name,
+                bindings,
+                callback: args,
```
CyrusNajmabadi: i'm a little confused that 'args' is both splatted with `...args` and passed to `callback`...  how does this work?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -415,6 +515,128 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
+     * The blob containing all the code for this FunctionApp.
+     */
+    public readonly zipBlob: storageForTypesOnly.ZipBlob;
+    /**
+     * The plan this Function App runs under.
+     */
+    public readonly plan: appservice.Plan;  
+    /**
+     * The plan this Function App runs under.
+     */
```
CyrusNajmabadi: wrong comment.

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: n/m.  i see what you're doing.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -415,6 +515,128 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
...
+    public readonly endpoint: pulumi.Output<string>;
+
+    constructor(type:string,
+                name: string,
+                args: ArchiveFunctionAppArgs,
+                opts: pulumi.ComponentResourceOptions = {}) {
+        super(type, name, undefined, opts);
+
+        const app = new CombinedFunctionApp(name, args, { parent: this });
```
CyrusNajmabadi: AFAICT, CombinedFunctionApp doesn't really serve much purpose.  It's just there to create all this underlying data, which you then export out of this type.  We technically could just use a normal non-resource object for CombinedFunctionApp.  Or just have a function that returned a pojo with this data.

CyrusNajmabadi: oh, i see.  you also have CallbackFuncitonApp extend this.  Ok. i need to write this all down :D

examples/http-external/Pulumi.yaml
```diff
@@ -0,0 +1,3 @@
+name: azure-http-external
+runtime: nodejs
+description: An http Azure Function App implemented in PowerShell
```
CyrusNajmabadi: can we add an example of a multi-function app as well?

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Yeah, it's a side effect of extending both `CallbackArgs<C, E, R>` and `FunctionAppArgsBase`, so those are split here.

examples/http-external/Pulumi.yaml
mikhailshilkov: Well, we don't have functions for that yet :)
We can add one function in the example itself for now, just to make sure API works.

mikhailshilkov: Added

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: I tried to outline this in https://github.com/pulumi/pulumi-azure/pull/264#issuecomment-501199117, let me know if you see better ways

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -415,6 +515,128 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
         this.container = container;
         this.plan = plan;
         this.zipBlob = zipBlob;
+
+        const routePrefix = args.hostSettings 
+            && args.hostSettings.extensions 
+            && args.hostSettings.extensions.http 
+            && args.hostSettings.extensions.http.routePrefix;
```
CyrusNajmabadi: can't wait for JS to get `.?` :)

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: 👍

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: 👍

PR #265, opened 2019-06-01
Switch upstream to terraform-providers/terraform-provider-azurerm

jen20:
This pull request switches the upstream of `pulumi-azure` from `pulumi/terraform-provider-azurerm` to `terraform-providers/terraform-provider-azurerm`. The only divergence in the fork was related to `ARM_PROVIDER_STRICT` which was enabled unconditionally.

We retain this behaviour by using `go:linkname`, as described in the block comment - repeated here:

The AzureRM Terraform Provider has a setting named ARM_PROVIDER_STRICT, the bulk of which was introduced into terraform-providers/terraform-provider-azurerm in commit 447abecad. The setting prevents accidental adoption of existing resources, and is something we want to opt into - not delegate to user control via the environment.

When we maintained our own fork of terraform-providers-azurerm, we just enabled the setting by modifying the feature flag value directly. Sadly it is unexported, and now we do _not_ depend on a fork but on upstream directly, we need a way to set that.

One option would be to use os.Setenv to set the environment variable for the child processes - but this would put Azure specific bits into the plugin loader which is not desirable. Instead, we use `go:linkname` to make the private package-level variable which controls this behaviour accessible to us.

go:linkname is not widely used outside the standard library, but allows us to effectively alias a particular symbol via an accessible name. Per Ian Lance Taylor's post on golang-dev [1], "it means: every time you are about to write out X in the object file, write out Y instead."

So, by aliasing a local variable `strictMode` to the particular symbol in question in the azure provider (identified via `nm`), we can set the private value here by assigning our own `strictMode` variable. Note that since we build with the vendor/ directory, we must use the qualified vendor path to the symbol.

Unfortunately, `go:linkname` does not give an error if the symbol being aliased does not exist. Since we are aliasing into someone elses package, there is a real risk that an upstream refactor will move or break this variable. To that end, as part of the lint step in Makefile, we verify that the built binary actually has the symbol listed here present. This does not protect against behavioural change, but we can likely write a test for the undesirable behaviour to validate against that.

Changes to this area of code are noted for upstream v1.30.0 (we are adopting this at v1.29.0) but do not affect the way we need to handle this.

[1]: https://groups.google.com/forum/#!topic/golang-dev/j2r7Vq6CBZk

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,7 @@ install_plugins::
 
 lint::
 	golangci-lint run
+	nm "$(shell which pulumi-resource-azure)" | grep github.com/pulumi/pulumi-azure/vendor/github.com/terraform-providers/terraform-provider-azurerm/azurerm.requireResourcesToBeImported
 
 install::
 	GOBIN=$(PULUMI_BIN) go install -ldflags "-X github.com/pulumi/pulumi-azure/pkg/version.Version=${VERSION}" ${PROJECT}/cmd/${PROVIDER}
--- a/cmd/pulumi-resource-azure/main.go
+++ b/cmd/pulumi-resource-azure/main.go
@@ -15,11 +15,56 @@
 package main
 
 import (
+	_ "unsafe" // Import go:linkname
+
+	"github.com/pulumi/pulumi-terraform/pkg/tfbridge"
+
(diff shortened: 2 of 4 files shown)


stack72:
So much happier with this approach so we don't need to maintain our own fork!

PR #274, opened 2019-06-07
Upgrade to v1.30.1 of Azurerm provider

stack72:
This is required due to a regression introduced in v1.30.0 upstream


stack72:
There are no code changes in this - it's only a dependency that needs upgraded due to an issue with authorization - I am going to self merge this because anyone using 0.18.6 will have failures in specific areas

Issue #232, opened 2019-05-07
Simplify output data flow in callback functions


mikhailshilkov:
## Problem

We've simplified the definition of Azure Functions so that people could use JS callbacks and deploy them as functions. The event data is provided as an input to the callback.

Typically, a function would need to produce some data and send it somewhere. At the moment, we only support one type of return data: HTTP response. Obviously, there are lots of other scenarios what developers might want to do. Some common examples:

- Send another message from a message-triggered function
- Save another blob from a blob-triggered function
- Save a document to Cosmos DB
- Enqueue a message from HTTP function

Say, I want to implement the last bullet point right now. Here is the code:

``` ts
import { Aborter, QueueURL, MessagesURL, ServiceURL, StorageURL, SharedKeyCredential } from "@azure/storage-queue";

const greeting = new azure.appservice.HttpEventSubscription('greeting', {
    resourceGroup,    
    callback: async (context, args) => {
        const account = storageAccount.name.get();
        const accountKey = storageAccount.primaryAccessKey.get();
        const sharedKeyCredential = new SharedKeyCredential(account, accountKey);
        const pipeline = StorageURL.newPipeline(sharedKeyCredential);
        const serviceURL = new ServiceURL(`https://${account}.queue.core.windows.net`, pipeline);
        const queueName = queue.name.get();
        const queueURL = QueueURL.fromServiceURL(serviceURL, queueName);
        const messagesURL = MessagesURL.fromQueueURL(queueURL);
        const contents = Buffer.from("Hello World!").toString('base64');
        const enqueueQueueResponse = await messagesURL.enqueue(Aborter.none, contents);
      return {
        status: 200,
        body: `Hello World!`
      };
    }
  });
```

This works but... doesn't look entirely clean.

## Approach 1: Helper functions

We could add some mixins to simplify the most typical scenarios and make examples/demos look very clean. For the above snippet, we could enable

``` ts
callback: async (context, args) => {
    const enqueueQueueResponse = await queue.messagesURL().enqueue(Aborter.none, "Hello World!");
    // or simply
    const enqueueQueueResponse = await queue.enqueue("Hello World!");
```

which would do the same but hide the mechanics.

**Pro's**: Clean. Uses simple code and minimum config.
**Con's**: Needs explicit work for each supported scenario. Needs to include and adapt specific SDKs for each sink type.

## Approach 2: Azure Functions output bindings

[...]

PR #275, opened 2019-06-07
 Support multiple bindings for CallbackFunctionApps

ChristianEder:
This PR adds the ability to extend CallbackFunctionApps with user-defined bindings - making them incredibly more useful by adding the ability to use any available input or output binding in addition to the binding(s) created by the mixins.

An example has been added to the examples/http-storage folder


ChristianEder:
@mikhailshilkov , @CyrusNajmabadi  Hi again :-)

ChristianEder:
@mikhailshilkov  I hope this doesn't conflict with your open PR - what you are doing over there is also something I'm waiting for :-)

ChristianEder:
Seems I unknowingly implemented one of the approaches discussed in #232 - but as I see now, the discussion there leaned towards the approach I chose (because as a Azure guy, not a pulumi guy, it was natural to do it that way), but wasn’t finalized yet.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -470,3 +474,14 @@ export function getResourceGroupNameAndLocation(
     const getResult = resourceGroupName.apply(n => core.getResourceGroup({ name: n }));
     return { resourceGroupName, location: getResult.location };
 }
+
+/** @internal */
+export function mergeBindings(additionalBindings: BindingDefinition[], bindings?: pulumi.Input<BindingDefinition[]>) : pulumi.Input<BindingDefinition[]>{
+    let merged : pulumi.Input<BindingDefinition[]>;
+    if (bindings) {
+        merged = pulumi.output(bindings!).apply(b => [...b, ...additionalBindings]);
```
CyrusNajmabadi: should need the `!` in `bindings!` here afaict.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -470,3 +474,14 @@ export function getResourceGroupNameAndLocation(
...
+
+/** @internal */
+export function mergeBindings(additionalBindings: BindingDefinition[], bindings?: pulumi.Input<BindingDefinition[]>) : pulumi.Input<BindingDefinition[]>{
+    let merged : pulumi.Input<BindingDefinition[]>;
+    if (bindings) {
+        merged = pulumi.output(bindings!).apply(b => [...b, ...additionalBindings]);
+    }else{
+        merged = additionalBindings;
+    }
+    return merged;
```
CyrusNajmabadi: consider rewriting this all as:

```ts
return bindings ? pulumi.output(bindings).apply(b => [...b, ...additionalBindings]) : additionalBindings;
```

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -89,6 +89,13 @@ export type CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> =
      */
     callbackFactory?: CallbackFactory<C, E, R>;
 
+    /**
+     * The bindings to be used for the function.
+     * See https://docs.microsoft.com/en-us/azure/azure-functions/functions-triggers-bindings 
+     * for more details
+     */
+    bindings?: pulumi.Input<azurefunctions.BindingDefinition[]>;
```
CyrusNajmabadi: ideall, this should be written as `pulumi.Input<pulumi.Input<azurefunctions.BindingDefinition>[]>`;

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -122,19 +122,21 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
+            type: "httpTrigger",
+            direction: "in",
+            name: "req",
+            route: args.route,
+            methods: args.methods,
+        }, {
+            type: "http",
+            direction: "out",
+            name: "$return",
+        }];
```
CyrusNajmabadi: in this case, can we indent one to help indicate these are children.  otherwise, it looks a bit wonky to me.

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -122,19 +122,21 @@ export class HttpEventSubscription extends mod.EventSubscription<mod.Context<Htt
...
+
+        args.bindings = mod.mergeBindings(bindings, args.bindings);
```
CyrusNajmabadi: i'd prefer we not mutate values flowing in.  so instead, just write this all as:

```ts
        super("azure:appservice:HttpEventSubscription", name, {
            ...args,
            bindings:  mod.mergeBindings(bindings, args.bindings),
```

sdk/nodejs/storage/zMixins.ts
```diff
@@ -228,16 +228,18 @@ export class BlobEventSubscription extends appservice.EventSubscription<BlobCont
...
 
+        args.bindings = appservice.mergeBindings(bindings, args.bindings);
+
         // Place the mapping from the well known key name to the storage account connection string in
         // the 'app settings' object.
         const account = pulumi.all([resourceGroupName, container.storageAccountName])
-                              .apply(([resourceGroupName, storageAccountName]) =>
-                                storage.getAccount({ resourceGroupName, name: storageAccountName }));
+                            .apply(([resourceGroupName, storageAccountName]) =>
+                              storage.getAccount({ resourceGroupName, name: storageAccountName }));
```
CyrusNajmabadi: can you undo the indentation change.  it was intentional to align the .all and .apply.  thanks!

sdk/nodejs/storage/zMixins.ts
```diff
@@ -248,6 +250,34 @@ export class BlobEventSubscription extends appservice.EventSubscription<BlobCont
...
+    /**
+     * The type of the table binding.  Must be 'table'.
+     */
+    type: "table";
+
+    /**
+     * The direction of the binding.  We only 'support' tables as outputs of functions.
+     */
+    direction: "out";
+
```
CyrusNajmabadi: extra newline.

sdk/nodejs/storage/zMixins.ts
```diff
@@ -411,20 +441,22 @@ export class QueueEventSubscription extends appservice.EventSubscription<QueueCo
...
         // the 'app settings' object.
         const appSettingsOutput = args.appSettings || pulumi.output({});
 
         // Place the mapping from the well known key name to the storage account connection string in
         // the 'app settings' object.
         const account = pulumi.all([resourceGroupName, queue.storageAccountName])
-                                .apply(([resourceGroupName, storageAccountName]) =>
-                                storage.getAccount({ resourceGroupName, name: storageAccountName }));
+                            .apply(([resourceGroupName, storageAccountName]) =>
+                            storage.getAccount({ resourceGroupName, name: storageAccountName }));
```
CyrusNajmabadi: same comment applies.

CyrusNajmabadi:
I very much like this.  Waiting on @mikhailshilkov  to make sure there's no concerns there.  This may end up conflicting with his in-flight PRs.  But i doubt it's something that we couldn't work through!

mikhailshilkov:
@ChristianEder Thanks for your idea, work, and PR. Before discussing the details, I'd love to settle on the public API space. My idea would be to try to make the API a bit more high-level, closer to what we do with triggers. The triggers don't expose raw bindings: instead, they provide abstractions like `onEvent`. I'm hoping to say something like `new TableOutputBinding(myTable)` and let it setup the binding, app settings, and what not.

@CyrusNajmabadi  I would also like to understand the outcome of #264 and how bindings would fit into "function objects" from there. I guess, the options would look similar in both cases. Or have you found another way?

ChristianEder:
I totally agree - the api discussion should be settled first, I just saw it after I had created the PR.

ChristianEder:
I will add some comments in #232 when I find the time

sdk/nodejs/appservice/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/appservice/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/appservice/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/appservice/zMixins_http.ts
ChristianEder: restored original indentation in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/appservice/zMixins_http.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/storage/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/storage/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

sdk/nodejs/storage/zMixins.ts
ChristianEder: fixed in d783c138f08bff57325be4e7fb98e3a6933d5b8b

ChristianEder:
@mikhailshilkov  you mean more like the example in 6702d4b943c719b054e7273f1cd20b2f7c262c91? Also see my comments in #232

examples/http-storage/index.ts
```diff
@@ -0,0 +1,38 @@
...
+      body: `Hello ${req.query.name}!`
+    };
+  },
+  methods: ["GET"],
+  account: storageAccount
+};
+
+storageAccount.bindTableOutput(args, "person", "Person");
+
+const greeting = new azure.appservice.HttpEventSubscription('greeting', args);
```
mikhailshilkov: I would try to keep the usual shape of API with args inline in the constructor call. The 3-step process feels a bit alien here. We should be able to do something like

``` ts
const greeting = new azure.appservice.HttpEventSubscription('greeting', {
  resourceGroup,
  callback: ...,
  methods: ["GET"],
  account: storageAccount,
  outputBindings: [storageAccount.tableOutput("person", "Person")]
});
```

with merging of app settings happening inside our implementation classes.

By the way, why not use an `azure.storage.Table` instance?

examples/http-storage/index.ts
```diff
@@ -0,0 +1,38 @@
...
+const args: azure.appservice.HttpEventSubscriptionArgs = {
+  resourceGroup,
+  callback: async (context, req: HttpRequest) => {
+
+    context.bindings.person = [];
+    context.bindings.person.push({
+      PartitionKey: req.query.email,
+      RowKey: req.query.email,
+      Name: req.query.name
+    });
```
mikhailshilkov: So, the [recommended](https://docs.microsoft.com/en-us/azure/azure-functions/functions-reference-node) way to use input and output bindings is to accept inputs as arguments and return outputs as properties:

``` ts
async function(context, myTrigger, myInput, myOtherInput) {
    let retMsg = `Hello, ${myInput} and ${myOtherInput}!`;
    return {
        httpResponse: {
            body: retMsg
        },
        queueOutput: retMsg
    };
}
```

I would very much like to encourage this style of callbacks and to support it as the default option. Ideally, we would find a TypeScript sorcery to check the number of arguments/returns versus the names and quantities of the bindings, but we may also do this at preview execution time.

examples/http-storage/index.ts
ChristianEder: What do you mean with 
>  inside our implementation classes

Do you mean that `HttpEventSubscription` or its base class implements logic that uses information from the `outputBindings` array of the args to modify both the app settings as well as the bindings used for the function? Sounds doable and I agree that this looks cleaner.

examples/http-storage/index.ts
ChristianEder: This comment basically applies only to the example, not the pulumi sdk part of the PR, right? I can change that, I just used a code sample that was on the docs (https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-table#output---javascript-example). I can fix this, though.

examples/http-storage/index.ts
mikhailshilkov: Yes, one of those (it's also changing a bit in the other PR which will hopefully be merged very soon).
I'll play with these ideas a bit more to shape the ideas more specifically, sorry for being vague.

ChristianEder: Since issue #232 is still open - is this now the way we want to go forward with this? If yes -> I'll refactor my PR to fit your proposal in a few days.

examples/http-storage/index.ts
mikhailshilkov: I suppose the current SDK won't let you change your example to mine: types define the callback signature quite strictly.
What's your opinion about which style of binding assignment is preferable?

examples/http-storage/index.ts
mikhailshilkov: Please check my latest comment in the issue https://github.com/pulumi/pulumi-azure/issues/232#issuecomment-501274842

ChristianEder: and regarding why I dont use `azure.storage.Table` : from the table, I dont have access to the connection string, but from the storage account, I do.

ChristianEder:
I'll close this PR, lets open a new one for the implementation when the other stuff has been merged

PR #280, opened 2019-06-13
Export Locations module for correct Location namespacing

stack72:


sdk/nodejs/location.ts
```diff
@@ -12,38 +12,40 @@
...
-export let UKSouth:            Location = "uksouth";
-export let UKWest:             Location = "ukwest";
-export let WestCentralUS:      Location = "westcentralus";
-export let KoreaCentral:       Location = "koreacentral";
-export let KoreaSouth:         Location = "koreasouth";
-export let FranceCentral:      Location = "francecentral";
-export let FranceSouth:        Location = "francesouth";
-export let SouthAfricaNorth:   Location = "southafricanorth";
-export let SouthAfricaWest:    Location = "southafricawest";
+export module Locations {
```
mikhailshilkov: Why capital `L`?

stack72: To go with the convention we have used elsewhere in other providers - @jen20 thoughts here?

stack72: @mikhailshilkov this comes from pulumi-aws#599 btw - I just spoke to @jen20 about it

jen20: The pattern comes from pulumi/pulumi-aws#599 - probably worth sticking with at this point?

PR #281, opened 2019-06-18
HTTP, Timer, and Queue functions

mikhailshilkov:
Introduce `HttpFunction`, `TimerFunction`, and `QueueFunction` - to be hosted within `MultiCallbackFunctionApp`. More functions to follow in a separate PR once we agree on the details of these three.

Two important changes that I made that we might want to iterate on:

1. `CallbackFunctionApp` and `EventSubscription` can accept a `Function` as a direct constructor parameter:

``` ts
constructor(type: string, name: string,
       bindings: pulumi.Input<BindingDefinition[]> | Function,
```

`bindings` name looks a bit ugly here, but the change allows reusing `XyzFunction` classes to pass everything needed down the callback app without duplication:

``` ts
super("azure:appservice:TimerSubscription", name, new TimerFunction(name, args), {
```

2. Added a dumb `FunctionBase` class which implements `Function` interface. Or should we design our Functions as factory functions instead of classes?

### DIFF
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -392,7 +392,7 @@ export interface Function {
     /**
      * Function callback.
      */
-    callback: CallbackArgs<any, any, any>;
+    callback: CallbackArgs<Context<any>, any, any>;
 
     /**
      * Application settings required by the function.
@@ -403,6 +403,42 @@ export interface Function {
 // Type guard for Function interface
 function isFunction(object: any): object is Function {
     return 'name' in object && 'bindings' in object && 'callback' in object;
+}
+
+/**
+ * Azure Function base class.
+ */
+export abstract class FunctionBase<C extends Context<R>, E, R extends Result> implements Function {
+    /**
+     * Function name.
+     */
+    public readonly name: string;
+
+    /**
+     * An array of function binding definitions.
+     */
+    public readonly bindings: pulumi.Input<BindingDefinition[]>;
+
+    /**
+     * Function callback.
+     */
+    public readonly callback: CallbackArgs<Context<any>, E, R>;
+
+    /**
+     * Application settings required by the function.
+     */
+    public readonly appSettings?: pulumi.Input<{ [key: string]: string }>;
+
+    constructor(name: string, 
+        trigger: pulumi.Input<BindingDefinition>,
+        inputOutputBindings: pulumi.Input<BindingDefinition[]>,
+        callback: CallbackArgs<C, E, R>,
+        appSettings?: pulumi.Input<{ [key: string]: string }>) {
+        this.name = name;
+        this.bindings = pulumi.all([trigger, inputOutputBindings]).apply(([t, iob]) => [t, ...iob]);
+        this.callback = <CallbackArgs<Context<any>, E, R>>callback;
+        this.appSettings = appSettings;
+    }
 }
 
 /**
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -405,21 +405,6 @@ function isFunction(object: any): object is Function {
     return 'name' in object && 'bindings' in object && 'callback' in object;
 }
 
-/** @internal */
-function makeFunction<C extends Context<R>, E, R extends Result>(
-    name: string, 
-    trigger: pulumi.Input<BindingDefinition>,
-    inputOutputBindings: pulumi.Input<BindingDefinition[]>,
-    callback: CallbackArgs<C, E, R>,
-    appSettings?: pulumi.Input<{ [key: string]: string }>): Function {
-    return {
-        name,
-        bindings: pulumi.all([trigger, inputOutputBindings]).apply(([t, iob]) => [t, ...iob]),
-        callback: <CallbackArgs<Context<any>, E, R>>callback,
-        appSettings,
-    };
-}
-
 /**
  * Azure Function base class.
  */
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -375,40 +375,9 @@ function redirectConsoleOutput<C extends Context<R>, E, R extends Result>(callba
 }
 
 /**
- * Interface to define individual Azure Functions. A function is not deployable by itself: instead, it must be passed
- * to MultiFunctionApp to be deployed as a part of the Azure Function App.
- */
-export interface Function {
-    /**
-     * Azure Function name.
-     */
-    name: string;
-
-    /**
-     * An array of function binding definitions.
-     */
-    bindings: pulumi.Input<BindingDefinition[]>;
-
-    /**
-     * Function callback.
-     */
-    callback: CallbackArgs<Context<any>, any, any>;
-
-    /**
-     * Application settings required by the function.
-     */
-    appSettings?: pulumi.Input<{ [key: string]: any }>;
-}
-
-// Type guard for Function interface
-function isFunction(object: any): object is Function {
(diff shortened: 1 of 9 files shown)


sdk/nodejs/appservice/zMixins.ts
```diff
@@ -400,6 +400,46 @@ export interface Function {
     appSettings?: pulumi.Input<{ [key: string]: any }>;
 }
 
+// Type guard for Function interface
+function isFunction(object: any): object is Function {
+    return 'name' in object && 'bindings' in object && 'callback' in object;
```
CyrusNajmabadi: i prefer the following:

`return obj && (<Function>obj).name && ...`

this way if we change anything about the shape of Function, we get an error here.  Right now, this is purely dynamic, so TS can't check anything.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -400,6 +400,46 @@ export interface Function {
     appSettings?: pulumi.Input<{ [key: string]: any }>;
 }
 
+// Type guard for Function interface
```
CyrusNajmabadi: FWIW, i think this comment is unnecessary :)

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -400,6 +400,46 @@ export interface Function {
...
 
+// Type guard for Function interface
+function isFunction(object: any): object is Function {
+    return 'name' in object && 'bindings' in object && 'callback' in object;
+}
+
+/**
+ * Azure Function base class.
+ */
+export abstract class FunctionBase<C extends Context<R>, E, R extends Result> implements Function {
```
CyrusNajmabadi: is there actual value in the Function interface?  Should we just have the Function base class?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -525,19 +565,14 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
      */
     public readonly endpoint: pulumi.Output<string>;
 
-    constructor(name: string, bindings: pulumi.Input<BindingDefinition[]>,
+    constructor(name: string, bindings: pulumi.Input<BindingDefinition[]> | Function,
```
CyrusNajmabadi: consider renamin to bindingsOrFunction

sdk/nodejs/appservice/zMixins_http.ts
```diff
@@ -61,6 +61,20 @@ export interface HttpHostSettings extends mod.HostSettings {
...
+    route?: pulumi.Input<string>;
+
+    /**
+     * An array of the HTTP methods to which the function responds. If not specified, the function
+     * responds to all HTTP methods.
+     */
+    methods?: pulumi.Input<pulumi.Input<string>[]>;
+}
+
 export type HttpEventSubscriptionArgs = util.Overwrite<mod.CallbackFunctionAppArgs<mod.Context<HttpResponse>, HttpRequest, HttpResponse>, {
```
CyrusNajmabadi: can we move away from .Overwrite here?  In our public APIs it becomes very confusing for people to know what's going on.

sdk/nodejs/appservice/zMixins_timer.ts
```diff
@@ -202,17 +214,7 @@ export type TimerSubscriptionArgs = util.Overwrite<mod.CallbackFunctionAppArgs<T
...
-     * A CRON expression for the timer schedule, e.g. '0 * * * * *'.
-     */
-    schedule: pulumi.Input<string | ScheduleArgs>;
-
-    /**
-     * If true, the function is invoked when the runtime starts.
-     */
-    runOnStartup?: pulumi.Input<boolean>;
-}>;
+} & TimerFunctionArgs>;
```
CyrusNajmabadi: same comment about Overwrite here.

CyrusNajmabadi:
Overall, likeing this direction a lot!

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Great, thanks for the hint!

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: It's the first time I write a guard, so I thought it's a big thing :) will remove

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: I tend to be cautious with classes, so I start with interfaces/functions, but then our current design seems to revolve around classes... Maybe we don't need both at this point...

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: isn't this a breaking change?

CyrusNajmabadi: nope.  names of parameters aren't aavailable in any meaningful form in js.  you can also write the above as:

```ts
constructor(name: string, bindings: pulumi.Input<BindingDefinition[]>);
constructor(name: string, func: Function);
constructor(name: string, bindingsOrFunc: pulumi.Input<BindingDefinition[]> | Function) { }
```

Though that's really up to you as the union type form is also ok.

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: understood.  in this case though we woudlnt' expect anyone to ever realy pass these in.  they'd only use the exact impls we provide.   so i think classes are fine here.

sdk/nodejs/appservice/zMixins_http.ts
mikhailshilkov: Do you mean `export interface HttpEventSubscriptionArgs extends mod.CallbackFunctionAppArgs<mod.Context<HttpResponse>, HttpRequest, HttpResponse>, HttpFunctionArgs` ?

How do we deal with mismatching types (`resourceGroupName`)?

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: There is a bit of distinction: the base class is generic - with types for the callback, while the interface is not. Non-generic interface is needed to have a collection of them, and the generic base class seems to make the end classes nicer. WDYT?

CyrusNajmabadi: Function can be a generic interface.  You can then have a colleciton of `Function<any, any, any>` to indicate that the specific type args aren't known.

mikhailshilkov: Great! Removed the interface, renamed the class to `Function`.

sdk/nodejs/appservice/zMixins_http.ts
CyrusNajmabadi: I basically mean: just write out the full interface that is expected (optionally extending from an existing interface if appropriate).  It may mean duplication.  But it is super simple to understand and document.  This is what we've done in awsx btw.

mikhailshilkov: Okay, it's all over mixins, let me take this as a separate PR

mikhailshilkov: #283

mikhailshilkov:
@CyrusNajmabadi Done with this round of changes. Do you want me to add more Functions to this PR or shall we merge this one and create a new one?

sdk/nodejs/appservice/zMixins_http.ts
CyrusNajmabadi: Sounds good. #resolved

PR #286, opened 2019-06-20
Create Function types for remaining triggers

### LINKED ISSUE #225: Multiple functions in a serverless app

mikhailshilkov:
Final part, fixes #225

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-___NULL___
+* Azure Function types for Blob, Service Bus, Event Hub, IoT Hub, and Cosmos DB triggers
+* BREAKING: Types related to ServiceBus Topic subscriptions: TopicContext, TopicHostSettings, and TopicCallback were replaced by more general types ServiceBusContext, ServiceBusHostSettings, and ServiceBusCallback to support queue-based scenarios.
 
 ---
 
--- a/examples/multi-callback-all/index.ts
+++ b/examples/multi-callback-all/index.ts
@@ -1,0 +1,165 @@
+import * as azure from "@pulumi/azure";
+
+const resourceGroup = new azure.core.ResourceGroup("example", { location: azure.Locations.WestUS2 });
+
+// HTTP
+const httpFunc = new azure.appservice.HttpFunction("http", {
+    callback: async (context, request) => ({ status: 200, body: "Hello World" }),
+});
+
+// Fire on schedule
+const timerFunc = new azure.appservice.TimerFunction("timer", {
+    schedule: { minute: 0 },
+    callback: async (context, timer) => {
+        console.log(`Running at ${timer}`);
+    },
+});
+
+// Storage Queue
+const storageAccount = new azure.storage.Account("storage", {
+    resourceGroupName: resourceGroup.name,
+    accountReplicationType: "LRS",
+    accountTier: "Standard",
+});
+
+const queue = new azure.storage.Queue("queue", {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+});
+
+const storageQueueFunc = new azure.storage.QueueFunction("storage-queue", {
+    queue,
+    callback: async (context, msg) => {
+        console.log(`Received a message ${msg.toString()}`);
+    },
+});
+
+// Storage Blobs
+const container = new azure.storage.Container("container", {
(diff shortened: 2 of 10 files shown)


CHANGELOG.md
```diff
@@ -3,7 +3,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-___NULL___
+* Azure Function types for Blob, Service Bus, Event Hub, IoT Hub, and Cosmos DB triggers
+* BREAKING: Types related to ServiceBus Topic subscriptions: TopicContext, TopicHostSettings, and TopicCallback were replaced by more general types ServiceBusContext, ServiceBusHostSettings, and ServiceBusCallback to support queue-based scenarios.
```
CyrusNajmabadi: any way to avoid the break?

CyrusNajmabadi:
I should be able to get to this in the next 24 hours.

CHANGELOG.md
mikhailshilkov: Yes, copy everything twice - keep `TopicXyz` and add `QueueXyz`, or use `TopicXyz` for queues which would look weird but will work.
Those 3 changed types aren't used explicitly in most cases, so I'm hoping it's not a big issue for users?

CyrusNajmabadi: Yeah, we added this recently enough that i'm ok with breaking.  Basically, it allows us to move fast, but quickly iterate if we want.  As things bake and become more solid we should be more careful about not breaking. but i think it's ok here as these are veyr very new.

examples/multi-callback-all/index.ts
```diff
@@ -0,0 +1,165 @@
...
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+});
+
+const storageQueueFunc = new azure.storage.QueueFunction("storage-queue", {
+    queue,
+    callback: async (context, msg) => {
+        console.log(`Received a message ${msg.toString()}`);
+    },
+});
```
CyrusNajmabadi: shoudl we consider makin it possible to get a QueueFunction directly from a Queue.  so like we have onEvent to make a FunctionApp, we could have getEventFunction to get this QueueFunction.  it would provide an easy way to discover this and to compse things easily.  Thoughts?

examples/multi-callback-all/index.ts
```diff
@@ -0,0 +1,165 @@
...
+    accountReplicationType: "LRS",
+    accountTier: "Standard",
+});
+
+const queue = new azure.storage.Queue("queue", {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+});
+
+const storageQueueFunc = new azure.storage.QueueFunction("storage-queue", {
```
CyrusNajmabadi: question for later, but what do we need this name for?

sdk/nodejs/cosmosdb/zMixins.ts
```diff
@@ -149,12 +158,26 @@ export class CosmosChangeFeedSubscription extends appservice.EventSubscription<C
...
 
-        const bindingConnectionKey = "BindingConnectionAppSettingsKey";
+        this.registerOutputs();
+    }
+}
+
+/**
+ * Azure Function triggered by a Cosmos DB Change Feed.
+ */
+export class CosmosDBFunction extends appservice.Function<CosmosChangeFeedContext, any[], void> {
```
CyrusNajmabadi: note to self: why does appservice.Function need a name?

examples/multi-callback-all/index.ts
mikhailshilkov: Each Azure Function has a name (visible in the portal, logs, management API, etc)

CyrusNajmabadi: yup.  LGTM .

sdk/nodejs/cosmosdb/zMixins.ts
CyrusNajmabadi: Resolved.

examples/multi-callback-all/index.ts
mikhailshilkov: Sounds promising! Seems non-invasive, so I'll give it a try in a separate PR.

PR #288, opened 2019-06-21
Introduce getEventFunction helpers

mikhailshilkov:
Make it possible to get e.g. a `QueueFunction` directly from a `Queue`. Like we have `onEvent` to make a FunctionApp, we now have `getEventFunction` to get the `QueueFunction`. It provides an easy way to discover this and to compose things easily.


ChristianEder:
So... onEvent() will put the function into its own function app while .getEventFunction() allows you to put it in your own function app, together with other functions? I think this should be clearer in the comments/docs, if I wouldn’t know, I would be confused by having the two. Besides that: nice work!

mikhailshilkov:
@ChristianEder Yeah, great point! Let's think about how to make this less ambiguous. Basically, we want to avoid people calling `getEventFunction` and then not using it anywhere assuming it's already wired.

ChristianEder:
I see two ways how to improve this:
- either update the comments to make it totally clear what either of the method does, also contrasting them against each other, and ideally also including links to the examples of either one
- or completely get rid of "onEvent": although its really neat & probably is an impressive demo of what can be done, in production apps its unlikely to deploy each function in a separate app.

Or: find better naming, but to be honest, this will be hard :-)

mikhailshilkov:
@ChristianEder I don't think we want/can remove `onEvent`. I've added more explicit comments to both functions, thanks for the suggestion!

CyrusNajmabadi:
I'm fine with just having good comments as to what's going on, thanks! :)

stack72:
@CyrusNajmabadi / @mikhailshilkov what's the situation with this PR - is it something we want to continue shaping?

mikhailshilkov:
@stack72 @CyrusNajmabadi I merged the latest changes in, so the PR is again ready for review

CyrusNajmabadi:
Nice!  This is awesome :)

PR #289, opened 2019-06-21
Input & Output bindings

mikhailshilkov:
Introduce the concepts of Input and Output bindings to all Azure Function mixins. Started with Storage Queue Outputs and Storage Table Inputs.


mikhailshilkov:
The feature is not complete yet, but I think it's worth running through a review. What's left to do:
- Enable input/output bindings for all other triggers in addition to HTTP and Storage
- Create input/output bindings for all destinations in addition to Storage (this is a potentially wider list than trigger types, so I think we can do that as separate PRs)

mikhailshilkov:
Also tagging @ChristianEder

examples/queue/index.ts
```diff
@@ -6,27 +6,68 @@ const resourceGroup = new azure.core.ResourceGroup("resourcegroup", {
...
+const queue2 = new azure.storage.Queue("queue2", {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+ });
+
+// HTTP Function will send a message to the first queue on each request
+const greeting = new azure.appservice.HttpEventSubscription('greeting', {
+    resourceGroup,
+    route: "{name}",
+    inputOutputs: [
```
ChristianEder: why `inputOutputs` and not `bindings`? Or 2 distinct properties, 1 for inputs, 1 for outputs?

mikhailshilkov: I had both of your versions in the past drafts :) 

I thought `bindings` *might* be confusing because they don't include the trigger (and the HTTP output), while technically a trigger is also a binding.

Two separate properties could make sense... I guess I'd need to have two distinct types then, to avoid misplacing?

ChristianEder: Your remark about `bindings` *might* be true. I personally would prefer to have 2 separate properties, but that’s not a strong opinion. If you choose to make it into two properties, you could probably relatively easy prevent misplacing by adding `& {direction:“out“}` to the type definition of outputs (and similar for inputs)

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -361,25 +361,57 @@ async function produceDeploymentArchiveAsync(args: MultiCallbackFunctionAppArgs)
...
+}
+
+/**
+ * Type alias for a response coming from an Azure Function callback, which applies to most Function types (HTTP being a notable exception).
+ */
+export type FunctionDefaultResponse =
+    // Nothing is returned if a function has no output bindings
+    void |
+    // A dictionary is returned if a function has output bindings
+    Record<string, any>;
```
CyrusNajmabadi: can we move this into the main description?  Also, can you doc what the keys/values are for this dictionary?

examples/queue/index.ts
CyrusNajmabadi: i think i woudl prefer two separate properties as well.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Done

examples/queue/index.ts
mikhailshilkov: Done, I like it too

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -758,3 +758,77 @@ export function combineBindingSettings(trigger: appservice.BindingSettings, inpu
...
+        if (body.masterKey === undefined || body.systemKeys === undefined || body.functionKeys === undefined) {
+            throw new Error(`Wrong shape of the host keys response: ${response.bodyAsText}`);
+        }
+
+        return body as FunctionHostKeys;
+    });
+};
+
+FunctionApp.prototype.getFunctionKeys = function(this: FunctionApp, functionName) {
+    return pulumi.all([this.id, functionName]).apply(async ([id, functionName]) => {
```
CyrusNajmabadi: thanks!

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -750,8 +779,10 @@ export function combineAppSettings(settings: pulumi.Input<{[key: string]: string
 }
 
 /** @internal */
-export function combineBindingSettings(trigger: appservice.BindingSettings, inputOutputs?: appservice.BindingSettings[]) {
-    const all = [trigger, ...inputOutputs || []];
+export function combineBindingSettings(trigger: BindingSettings<BindingDefinition>,
+                                       inputs?: InputBindingSettings[],
+                                       outputs?: OutputBindingSettings[]) {
+    const all = [trigger, ...inputs || [], ...outputs || []];
```
CyrusNajmabadi: oh wow.   i didn't know that works.  looks like that parses as `...(inputs || [])`.

examples/blob/index.ts
```diff
@@ -44,7 +44,7 @@ const sampleFile = new azure.storage.Blob("test.html", {
 const fileServer = new azure.appservice.HttpEventSubscription("file-server", {
     resourceGroup,
     route: "{name}",
-    inputOutputs: [
+    inputs: [
         storageContainer.input("blob", { blobName: "{name}" }),
     ],
```
CyrusNajmabadi: i far prefer this :)

CyrusNajmabadi:
This is awesome.  Sorry it took so long to get in!

PR #291, opened 2019-06-24
Adding constants for Azure Storage Account Tier, Kinds, Types and Replication Types

stack72:


sdk/nodejs/storage/accessTier.ts
```diff
@@ -0,0 +1,22 @@
...
+// limitations under the License.
+
+export module AccessTiers {
+    export const BlobStorage: AccessTier = "BlobStorage";
+    export const StorageV2:   AccessTier = "StorageV2";
+}
+
+export type AccessTier =
+    "BlobStorage" |
+    "StorageV2"   ;
```
mikhailshilkov: AFAIK `Access Tier` is something like `Hot` or `Cold`

sdk/nodejs/storage/account.ts
```diff
@@ -430,7 +435,7 @@ export interface AccountState {
     /**
      * Defines the access tier for `BlobStorage` and `StorageV2` accounts. Valid options are `Hot` and `Cool`, defaults to `Hot`.
      */
-    readonly accessTier?: pulumi.Input<string>;
+    readonly accessTier?: pulumi.Input<string | AccessTier>;
```
mikhailshilkov: Wouldn't we be able to make it simply `pulumi.Input<AccessTier>` (applies to all types)? I think these values are less ambiguous than locations, and we may just enforce the proper values on users?

examples/blob/index.ts
```diff
@@ -3,14 +3,14 @@
...
+    location: azure.Locations.WestUS2,
 });
 
 // Create a storage account for our images
 const storageAccount = new azure.storage.Account("storage", {
     resourceGroupName: resourceGroup.name,
-    accountReplicationType: "LRS",
-    accountTier: "Standard",
+    accountReplicationType: azure.storage.AccountReplicationTypes.LRS,
+    accountTier: azure.storage.AccountTiers.Standard,
```
mikhailshilkov: That's a lot more characters than before. If we can enforce the values in types (see my other comment), wouldn't it be okay to keep "LRS" and "Standard"? people would still get a compile error if they mistype it.
Might be a subjective choice - but we'd need to pick a better one for all the examples.

sdk/nodejs/storage/account.ts
stack72: I just didn't want to make a breaking change

sdk/nodejs/storage/accessTier.ts
stack72: I took these directly from the Terraform provider docs...

mikhailshilkov: They got you tricked: 
```
Defines the access tier for `BlobStorage` and `StorageV2` accounts. 
Valid options are `Hot` and `Cool`, defaults to `Hot`.
```

sdk/nodejs/storage/account.ts
mikhailshilkov: I think we discussed with @jen20 that this kind of change might be fine, let's wait for his take

stack72:
I am going to close this out - we can come back to this another time and work out what we need to do here

Issue #277, opened 2019-06-09
Create an AzureX package


joeduffy:

stack72:
@CyrusNajmabadi / @joeduffy this is in progress now - the repo has been created so I am going to close this out

PR #293, opened 2019-06-27
Add support for dashboards

ChristianEder:
Added the capability to deploy Azure dashboards like the one seen below.

This PR is still in progress, but would like to get your feedback early. The most important points that I currently see are:

- I am experiencing weird behavior with the new zMixins.ts file I added in this PR to the appinsights module, perhaps you guys can give me a hint, why my mixin is not applied / compiled when I dont actually export something out of the mixin AND use that export in my example --> see "examples\dashboard\parts\requestAnalytics\part.ts" line 5
- I want to add built-in support for more types of tiles / parts
- The syntaxt to define the donut chart (examples\dashboard\parts\requestAnalytics\part.ts) is pretty weird, and I would like to abstract that away somehow, but on the other hand its just like how Azure dashboards are... any ideas? 

image


mikhailshilkov:
Thanks for putting this together! Interesting stuff! I'd definitely want to support more operational scenarios, including the dashboards.

Have you considered creating a [dynamic provider](https://pulumi.io/reference/programming-model/#dynamicproviders) for them instead of the mixins? This could have a couple of benefits:

- Supporting multiple client languages (js/ts + python, go).
- Being self-contained, so easy to opt-in and opt-out

I may be wrong but mixins shine when you need to use some TypeScript-specific language features to improve the developer experience (like, callback serialization for serverless functions). @CyrusNajmabadi @lukehoban ?

ChristianEder:
Hmm I didnt know about those. However, the docs state that dynamic providers are more geared towards scenarios where you as a pulumi user want to create your own resources not supported by pulumi yet

> Define an implementation of the Provider interface directly within your Pulumi program, just for use within that program.

This sound like that if I wanted to add dashboard support using dynamic providers (and want the benefit of supporting all supported languages), I'd have to implement them separately in each of the languages, and distribute them as a seperate npm package (or the go and python equivalents of that)?

mikhailshilkov:
> I'd have to implement them separately in each of the languages

I had an impression that one could package a dynamic provider as a normal provider, but I guess I got confused with that.

By the way, you can find examples of dynamic providers [here](https://github.com/pulumi/examples/blob/master/azure-ts-static-website/staticWebsite.ts) and [here](https://github.com/pulumi/examples/blob/master/azure-ts-dynamicresource/cdnCustomDomain.ts).

ChristianEder:
So nevertheless you propose to implement dashboards via dynamic provider? In this case, i habe a few questions:

- the Dynamic resource & Provider will get packaged and distributed in its own npm package separate form @pulumu/azure?

- is there a „dynamic resource“ way of adding functionality to other already existing resources, such as factory functions at e.g. the app insights resource to create dashboard parts form there as seen in the example I added? I assume: the mixin thing will work in this kind of setup the same way, although the mixin is in a separate package

- how do dynamic resource packages get discovered, aka how would anybody know that there is a package that implements dashboards?

- all the samples you linked and that i found in the docs are either using Rest calls or command line tools to do their things. Could the provider still use the existing template Deployment resource to create the dashboard?

mikhailshilkov:
I think I confused you and myself with my comment about the dynamic provider. Sorry for that! Let's start again :)

Basically, we have two questions:

- How do we package the code
- How do we represent the dashboard as a pulumi resource

We should think of what happens if Terraform introduces their own Dashboard and we inherit it from the generated code. To me, your code could be part of [azurex](https://github.com/pulumi/pulumi-azure/issues/277) package, which doesn't exist yet. As I don't know the future direction, I'll leave this to others to comment.

In any case, I would love to see the dashboard as a first-class resource. Right now, it's a `TemplateDeployment`:

```
     Type                                       Name                 Plan
 +   pulumi:pulumi:Stack                        azure-dashboard-dev  create
 +   ├─ azure:appservice:HttpEventSubscription  greeting             create
...
 +   ├─ azure:core:ResourceGroup                example              create
 +   ├─ azure:appinsights:Insights              insights             create
 +   └─ azure:core:TemplateDeployment           dashboard            create
```

So, if I ask for details, I see the JSON. If I change the dashboard, the diff is quite limited:

```
     Type                              Name             Plan       Info
     pulumi:pulumi:Stack               azure-dashboard            
 ~   └─ azure:core:TemplateDeployment  dashboard        update     [diff: ~parameters,templateBody]
```

IMO, it would be great to see a Dashboard-specific type as the top-level resource with some identifyable properties. The good news is that you probably don't need a dynamic resource for that, just inheriting from `CustomResource` or `ComponentResource` and defining the properties.

Does this make sense to you? Once again, sorry for the confusion, I'm still learning too :)

ChristianEder:
This does actually make more sense to me than the dynamic resource thing - but since I never wrote a CustomResource before, do you have some example to point me to? I find a lot of CustomResource implementations in the pulumi azure codebase, but they are auto-generated, not handwritten. Also: does this solution give an answer to the question what will happen when TF supports dashboards natively?

mikhailshilkov:
> do you have some example to point me to

No, me neither. That's why I started with `DynamicProvider` - at least I have experience and examples in building those. Maybe `ComponentResource` would be enough - I'd need to try it out.

> does this solution give an answer to the question what will happen when TF supports dashboards natively?

Not really: it's unlikely that our today's API would be compatible with the hypothetical Terraform API.

ChristianEder:
I just checked - it seems that even the az CLI only supports creating dashboards via ARM templates. So - I'll try to wrap my call to the TemplateDeployment constructor into a dynamic or ComponentResource, plus expose some details of the dashboard as outputs?

ChristianEder:
Hmm - I now refactored Dashboard, Lense, Part to `ComponentResource`, but still I dont see a more detailed diff when I change something and run `pulumi up`

mikhailshilkov:
@ChristianEder Right... It seems I keep confusing you, sorry about that!

@CyrusNajmabadi Could you have a look at this PR and advise the best way forward?

CyrusNajmabadi:
Hi there!  Open questions i have:

1. have you taken a look at how we do cloudwatch dashboards in awsx?  If possible, it would be great if we can follow the same sorts of patterns here (though things obviously have to be different as these are totally different systems).  However, in terms of easily specifying widgets and having us take care of htings, it would be great to have a consistent theme across both.
2. This feels like this shoudl be in azurex.  Does it have to be in azure?  If not, this can motivate us actually going and building that package to start putting the right things into it.

ChristianEder:
Hi @CyrusNajmabadi : 

- Regarding the AWS CloudWatch remark: no I haven't, fair point, will do.
- Regarding azurex: It totally hasn't have to be in the azure package. How will the azurex package be set up and by whom?

mikhailshilkov:
@ChristianEder Azure X would be a new repo under Pulumi org, here is the open issue https://github.com/pulumi/pulumi-azure/issues/277 assigned to @CyrusNajmabadi

CyrusNajmabadi:
Here's how i did the cloudwatch dashboards: https://github.com/pulumi/pulumi-awsx/tree/master/nodejs/awsx/cloudwatch

There's an explanation here: 

https://github.com/pulumi/pulumi-awsx/tree/master/nodejs/awsx/cloudwatch#dashboards

And interesting code is in all the dashboard.ts and widget_xxx.ts

CyrusNajmabadi:
> @ChristianEder Azure X would be a new repo under Pulumi org, here is the open issue #277 assigned to @CyrusNajmabadi

Indeed :)

If you're passionate about persuing this, I can make azurex happen.  We'd likely start it out as a playground for interesting ideas to be experimented with.  Then we can graduate-up things we like to be fully supported modules.  We did this initially with awsx.  Let me know if you'd like me to do that!

mikhailshilkov:
I would go ahead and create it so that we could also start collecting ideas for what could be valuable additions to `azurex` in the issues.

ChristianEder:
Great - let me know when its there, so i can move my PR

CyrusNajmabadi:
Yup yup!  Should be able to do this today.

CyrusNajmabadi:
@ellismg Is going to set this up for you guys!

ChristianEder:
Great, thx for doing that! Anyone already has a plan how that repository will be structured?

CyrusNajmabadi:
@ChristianEder I would ideally structure like awsx.  We could literally populate it with the same-ish folder structure, and same core files (i.e. those at the root, makefiles, etc.).  Except, obviously, references to AWS would be Azure :)

CyrusNajmabadi:
I can do this this weekend if you want :)  Or, if you feel up to populating, you should feel free :)

ChristianEder:
Sorry, I’ve got a small kid at home -  there is no such thing as this “weekend” you talk about :-)

ChristianEder:
Closing this in favor of a new one I am going to open on azurex

Issue #295, opened 2019-07-02
Support EventGrid triggers in callback functions


mikhailshilkov:
I am spinning this out of #218 because EventGrid is a bit different from other triggers and warrants some extra considerations on the design of functions.

#### Azure Event Grid

[Azure Event Grid](https://azure.microsoft.com/en-us/services/event-grid/) is the latest messaging service in Azure and is positioned as a massively-scalable event routing service to power event-driven and serverless apps. Event Grid subscriptions are basically webhooks, where an HTTP request is sent to the specified endpoint for each event in a given topic.

One of the unique features of Event Grid is the existence of many built-in topics with events from native Azure services. For instance, each Storage Account has an endpoint to subscribe for events in that account: e.g., a call would be made every time a new blob is uploaded. Similar endpoints exist for events from subscription, resource groups, container registries, media services, maps, and so on, increasing over time.

#### Event Grid trigger

Azure Functions have a dedicated Event Grid trigger. Effectively, it's an HTTP function with some extra wiring to handle authentication hand-shake initiated by Event Grid.

The trigger [definition](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-event-grid#javascript-example) is very simple:

```
{
    "type": "eventGridTrigger",
    "name": "eventGridEvent",
    "direction": "in"
}
```

The function binding does not define the events to subscribe to. It's the Event Grid subscription which makes it all work.

#### Event Grid subscription

Event subscription is mainly a combination of three things:

- Event Grid scope (the broad "topic" to subscribe to, the event source)
- Webhook URL (A function webhook URL)
- Optional filter to narrow the type of events to receive

Multiple subscriptions can be created for the same scope and/or for the same webhook (Azure Function).

#### 1. Custom topic callback

One can define a custom Event Grid topic and subscribe on it with a callback. This looks very similar to our usual scenario of callback functions, so the code could be:

``` ts
const topic = new azure.eventhub.EventGridTopic("bla", {
    resourceGroupName: resourceGroup.name,
});

eg.onEvent("onBla", async (context, event) => {
    console.log(JSON.stringify(context));
    console.log(JSON.stringify(event));
});
```

#### 2.1. Built-in topic subscriptions

A major, and arguably even more useful, use case is subscribing to built-in events of Azure. For instance, to get events from a storage account:

``` ts
const storageAccount = new azure.storage.Account("play", {
    resourceGroupName: resourceGroup.name,
    accountReplicationType: "LRS",
[...]

PR #297, opened 2019-07-08
Add functions to list keys of Function Apps

mikhailshilkov:
Adds mixin functions to wrap ARM API calls to List Keys of Function Apps, as described in [this issue](https://github.com/Azure/azure-functions-host/issues/3994).

Required for #295 or in case we want to support non-anonymous HTTP callback functions.

A function `getServiceClientCredentials` got introduced in `core` package. It might be useful for any REST interaction with the Azure Management API.

I enabled "format on save" in my VS Code, so some old lines got re-formated. Let me know if that's too annoying.

### DIFF
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -15,6 +15,7 @@
 import * as pulumi from "@pulumi/pulumi";
 
 import * as azurefunctions from "@azure/functions";
+import { AzureServiceClient } from "@azure/ms-rest-azure-js";
 
 import { FunctionApp } from "./functionApp";
 
@@ -196,7 +197,7 @@ interface FunctionAppArgsBase {
     /**
      * A mapping of tags to assign to the resource.
      */
-    readonly tags?: pulumi.Input<{[key: string]: any}>;
+    readonly tags?: pulumi.Input<{ [key: string]: any }>;
 
     /**
      * The runtime version associated with the Function App. Defaults to `~2`.
@@ -248,7 +249,7 @@ export interface HostSettings {
          * A sliding time window used in conjunction with the `healthCheckThreshold` setting.
          * Defaults to 2 minutes.
          */
-        healthCheckWindow:string,
+        healthCheckWindow: string,
         /**
          * Maximum number of times the health check can fail before a host recycle is initiated.  Defaults to `6`.
          */
@@ -355,13 +356,13 @@ async function produceDeploymentArchiveAsync(args: MultiCallbackFunctionAppArgs)
         const body = await serializeFunctionCallback(func.callback);
 
         map[`${func.name}/index.js`] = new pulumi.asset.StringAsset(`module.exports = require("./handler").handler`),
-        map[`${func.name}/handler.js`] = new pulumi.asset.StringAsset(body.text);
+            map[`${func.name}/handler.js`] = new pulumi.asset.StringAsset(body.text);
     }
 
     return new pulumi.asset.AssetArchive(map);
 }
 
-function combineAppSettings(args: MultiCallbackFunctionAppArgs): pulumi.Output<{[key: string]: string}> {
+function combineAppSettings(args: MultiCallbackFunctionAppArgs): pulumi.Output<{ [key: string]: string }> {
     const applicationSetting = args.appSettings || {};
     const perFunctionSettings = args.functions !== undefined ? args.functions.map(c => c.appSettings || {}) : [];
     return pulumi.all([applicationSetting, ...perFunctionSettings]).apply(items => items.reduce((a, b) => ({ ...a, ...b }), {}));
@@ -436,9 +437,9 @@ export interface ArchiveFunctionAppArgs extends FunctionAppArgsBase {
 };
 
 function createFunctionAppParts(
-        name: string,
-        args: ArchiveFunctionAppArgs,
-        opts: pulumi.CustomResourceOptions = {}) {
+    name: string,
+    args: ArchiveFunctionAppArgs,
+    opts: pulumi.CustomResourceOptions = {}) {
 
     if (!args.archive) {
         throw new Error("Deployment [archive] must be provided.");
@@ -540,7 +541,7 @@ export class CallbackFunctionApp<C extends Context<R>, E, R extends Result> exte
     public readonly endpoint: pulumi.Output<string>;
 
     constructor(name: string, bindingsOrFunc: pulumi.Input<BindingDefinition[]> | Function<C, E, R>,
-                args: CallbackFunctionAppArgs<C, E, R>, opts: pulumi.CustomResourceOptions = {}) {
+        args: CallbackFunctionAppArgs<C, E, R>, opts: pulumi.CustomResourceOptions = {}) {
 
         const functions = bindingsOrFunc instanceof Function ? [bindingsOrFunc] : [<Function<C, E, R>>{ name, bindings: bindingsOrFunc, callback: args }];
         const parts = createFunctionAppParts(name, {
@@ -594,10 +595,10 @@ export abstract class PackagedFunctionApp extends pulumi.ComponentResource {
      */
     public readonly endpoint: pulumi.Output<string>;
 
-    constructor(type:string,
-                name: string,
-                args: ArchiveFunctionAppArgs,
-                opts: pulumi.ComponentResourceOptions = {}) {
+    constructor(type: string,
+        name: string,
+        args: ArchiveFunctionAppArgs,
+        opts: pulumi.ComponentResourceOptions = {}) {
         super(type, name, undefined, opts);
 
         const parentOpts = { parent: this };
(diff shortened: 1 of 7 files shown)


mikhailshilkov:
Note that we might get this Terraform on some point, but they are blocked upstream, so probably not in short-term. See https://github.com/terraform-providers/terraform-provider-azurerm/issues/699

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -355,13 +356,13 @@ async function produceDeploymentArchiveAsync(args: MultiCallbackFunctionAppArgs)
         const body = await serializeFunctionCallback(func.callback);
 
         map[`${func.name}/index.js`] = new pulumi.asset.StringAsset(`module.exports = require("./handler").handler`),
-        map[`${func.name}/handler.js`] = new pulumi.asset.StringAsset(body.text);
+            map[`${func.name}/handler.js`] = new pulumi.asset.StringAsset(body.text);
```
CyrusNajmabadi: undo?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -436,9 +437,9 @@ export interface ArchiveFunctionAppArgs extends FunctionAppArgsBase {
 };
 
 function createFunctionAppParts(
-        name: string,
-        args: ArchiveFunctionAppArgs,
-        opts: pulumi.CustomResourceOptions = {}) {
+    name: string,
+    args: ArchiveFunctionAppArgs,
+    opts: pulumi.CustomResourceOptions = {}) {
```
CyrusNajmabadi: i prefer the previous indentation because it makes it clear these are wrapped params, and not the starts of statements.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: that's because of the comma on the previous line. remove that one?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -594,10 +595,10 @@ export abstract class PackagedFunctionApp extends pulumi.ComponentResource {
...
     public readonly endpoint: pulumi.Output<string>;
 
-    constructor(type:string,
-                name: string,
-                args: ArchiveFunctionAppArgs,
-                opts: pulumi.ComponentResourceOptions = {}) {
+    constructor(type: string,
+        name: string,
+        args: ArchiveFunctionAppArgs,
+        opts: pulumi.ComponentResourceOptions = {}) {
```
CyrusNajmabadi: undo please

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -711,3 +712,57 @@ export function getResourceGroupNameAndLocation(
...
+    default: string;
+    [key: string]: string;
+}
+
+declare module "./functionApp" {
+    interface FunctionApp {
+        /**
+         * Retrieve the keys associated with the Function App.
+         */
+        listHostKeys(): pulumi.Output<FunctionHostKeys>;
```
CyrusNajmabadi: 'list' seems odd.  why not just 'get'?

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -711,3 +712,57 @@ export function getResourceGroupNameAndLocation(
...
+    }
+}
+
+FunctionApp.prototype.listHostKeys = function (this: FunctionApp) {
+    return this.id.apply(async id => {
+        const credentials = await core.getServiceClientCredentials();
+        const client = new AzureServiceClient(credentials);
+        const url = `https://management.azure.com${id}/host/default/listkeys?api-version=2018-02-01`;
+        const response = await client.sendRequest({ method: "POST", url });
+        return <FunctionHostKeys>response.parsedBody;
```
CyrusNajmabadi: 1. should we be chekcing the response for things like errors?
2. is there no actual API call we can make here that's more strongly typed than just hittin a random https url

sdk/nodejs/core/zMixins.ts
```diff
@@ -0,0 +1,37 @@
...
+export async function getServiceClientCredentials(): Promise<ServiceClientCredentials> {
+    let credentials: ServiceClientCredentials;
+
+    if (config.useMsi) {
+        credentials = await msnodeauth.loginWithAppServiceMSI({ msiEndpoint: config.msiEndpoint });
+    } else if (config.clientId && config.clientSecret && config.tenantId) {
+        credentials = await msnodeauth.loginWithServicePrincipalSecret(config.clientId, config.clientSecret, config.tenantId);
+    } else {
+        // `create()` will throw an error if the Az CLI is not installed or `az login` has never been run.
+        const cliCredentials = await msnodeauth.AzureCliCredentials.create();
```
CyrusNajmabadi: this variable seems unnecessary.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: That's the term they use, might make it a bit better discoverable? Otherwise, agree.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: 1. Yes, but I guess we can only wrap change the error that we throw?
2. Not that I know of, they've just released the API itself a few weeks ago.

sdk/nodejs/appservice/zMixins.ts
mikhailshilkov: Renamed

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: yeah, that comma should be a semicolon :)

sdk/nodejs/appservice/zMixins.ts
CyrusNajmabadi: ok.  i'm ok either way.  i leave it to you what you feel is best.  i personally think 'get' is totally fine.

sdk/nodejs/appservice/zMixins.ts
```diff
@@ -711,3 +711,77 @@ export function getResourceGroupNameAndLocation(
...
+        if (body.masterKey === undefined || body.systemKeys === undefined || body.functionKeys === undefined) {
+            throw new Error(`Wrong shape of the host keys response: ${response.bodyAsText}`);
+        }
+
+        return body as FunctionHostKeys;
+    });
+};
+
+FunctionApp.prototype.getFunctionKeys = function(this: FunctionApp, functionName) {
+    return pulumi.all([this.id, functionName]).apply(async ([id, name]) => {
```
CyrusNajmabadi: i'm feeling dumb, but i cannot understand what this functionName is for.  is it just to ensure that hte function has been created first before getFunctionKeys has been called?  it seems weird...

CyrusNajmabadi:
Done with review pass.  LGTM, but i did hav ea question about something odd...

sdk/nodejs/appservice/zMixins.ts
ChristianEder: Its used below in line 773 in the API call to the ARM Api to identify the function inside of the function app.

mikhailshilkov: Each function has its own set of keys in addition to the host keys, so we request URL contains the name as @ChristianEder kindly noticed

CyrusNajmabadi: oh... doh.  I see now.  i was looking for functionName, and not 'name'.  In general we actually *like* reusing the names (i.e. `pulumi.all([this.id, functionName]).apply(async ([id, functionName])`.  It helps indicate they represent the same thing, and it makes sure that the outer variable is blocked from being used in the inner scope.  I was looking for that pattern and that's what threw me off.  This is all good, thanks for clarifying!

mikhailshilkov: That makes sense, I'll rename and pick up this pattern.

sdk/nodejs/core/zMixins.ts
```diff
@@ -0,0 +1,37 @@
...
+
+    if (config.useMsi) {
+        credentials = await msnodeauth.loginWithAppServiceMSI({ msiEndpoint: config.msiEndpoint });
+    } else if (config.clientId && config.clientSecret && config.tenantId) {
+        credentials = await msnodeauth.loginWithServicePrincipalSecret(
+            config.clientId, config.clientSecret, config.tenantId);
+    } else {
+        // `create()` will throw an error if the Az CLI is not installed or `az login` has never been run.
+        credentials = await msnodeauth.AzureCliCredentials.create();
+    }
```
lukehoban: Are we sure these are the complete set of configuration options supported by the Azure provider?  I'm a little nervous re-creating this logic outside of the provider.  Could we perhaps cross-reference to the primary source for this logic? (Ideally in the backing provider to make sure it's really the same as what the user is seeing for other `@pulumi/azure` operations.

mikhailshilkov: Yeah, me too... I don't know of the source of truth for this.
Ideally, I'd use some function provided by Pulumi core/Terraform which would give me an access token that the provider is already using. Is there something like that? I asked on Slack once but didn't get an answer.

PR #300, opened 2019-07-09
Allow specifying the data type of a Queue Function message

### LINKED ISSUE #290: Stop using dataType=binary for queue-triggered subscriptions (functions)?
Our `QueueEventSubscription` and `QueueFunction` use the binding which defines 

```
/**
 * How we want the message represented when passed into the callback. We specify 'binary'
 * so that all data is passed in as a buffer.  Otherwise, Azure will attempt to sniff
 * the content and convert it accordingly. This gives us a consistent way to know what
 * data will be passed into the function.
 */
dataType: "binary";
```

I think this might be a wrong configuration for several reasons:

1. [The official guide](https://docs.microsoft.com/en-us/azure/azure-functions/functions-bindings-storage-queue#trigger---configuration) doesn't mention `dataType`. I suspect we copied it from Blob Trigger where it is documented. However, it does work for queues too.

2. As a consequence of the configuration, we use `Buffer` as the input type for callbacks. That means that a message `{ "name": "John" }` needs to be deserialized in every callback: `JSON.parse(buffer.toString())`. While `Buffer` may be okay as an advanced option, I estimate that most people just send JSON or strings over the wire, so their life gets a bit more complicated.

3. Azure Functions have a feature called input bindings. I'm working on supporting it in #289. An example could look like this:

    ``` ts
    queue1.onEvent("NewMessage",  {
        inputOutputs: [
            rolesTable.input("tableIn", { filter: "Name eq '{name}'"})
        ],
        callback: async (context, person, roles: any[]) => {
            // roles contains data from the table storage looked up by name
        },
    });
    ```

    Note how the input binding data lookup is bound to the `name` property of the incoming message. This works if the data type is not binary (i.e. `person` is just a JS object). The whole input binding expression feature doesn't work for `dataType: binary`.

4. `dataType: binary` doesn't seem to work for the output binding, so it's actually hard to *send* a `Buffer`. (I'd need to try again if we want to be 100% sure)

5. Our implementation of Service Bus queues, topics, and Event Hubs doesn't use `binary`.

We could just remove `dataType: binary` from the queue triggers. Obviously, that's a breaking change. 

Instead, we could introduce an option to configure it. Do we have to leave it as default (which might complicate life before people discover it)? What do we do with `Buffer` type currently used everywhere? Replace it with `any`?

@CyrusNajmabadi thoughts?

mikhailshilkov:
- Introduce `dataType` option for Storage Queue based callback functions
- Set the default to data type `string`

It's a breaking change.

Fixed #290


sdk/nodejs/storage/zMixins.ts
```diff
@@ -321,39 +319,46 @@ export interface QueueContext extends appservice.Context<void> {
...
 }
 
 /**
  * Signature of the callback that can receive queue notifications.
  */
-export type QueueCallback = appservice.Callback<QueueContext, Buffer, void>;
+export type QueueCallback = appservice.Callback<QueueContext, any, void>;
 
-export interface QueueFunctionArgs extends appservice.CallbackArgs<QueueContext, Buffer, void> {
+export interface QueueFunctionArgs extends appservice.CallbackArgs<QueueContext, any, void> {
```
CyrusNajmabadi: should this be typed as `Buffer | string | object`?

sdk/nodejs/storage/zMixins.ts
```diff
@@ -371,6 +376,13 @@ export interface QueueEventSubscriptionArgs extends appservice.CallbackFunctionA
      * be used in their place.
      */
     hostSettings?: QueueHostSettings;
+
+    /**
+     * Specify 'string' to get messages as a string, or a parsed object if the message is a valid JSON literal.
+     * Specify 'binary' for all data to be passed in as a Buffer.
+     * Default is 'string'.
+     */
+    dataType?: "binary" | "string";
```
CyrusNajmabadi: shouldn't this be `pulumi.Input<...>`?

CyrusNajmabadi:
Small nits and questions.  :)

sdk/nodejs/storage/zMixins.ts
mikhailshilkov: how is this helpful? (honest question)

sdk/nodejs/storage/zMixins.ts
mikhailshilkov: We could... Not all arguments are `Input`s and I can't think of a good example when this would be useful here. But we could.

sdk/nodejs/storage/zMixins.ts
CyrusNajmabadi: it makes the callback clearer about the tyeps of data you might expect to get.  i.e. `any` literally means: you could get anything back (like a boolean), at least with the union type we're being explicit that you should at least expect to get one of these types back.

Just a thought :)

mikhailshilkov: So, I gave it a try: sending a message "true" results in a `boolean` argument (printed `typeof`), "1" becomes a `number`. They are valid JSON...

CyrusNajmabadi: Ahhhh.  Fascinating!  Ok `any` it is.  We might want to mention that JSON behavior more explicitly (i.e. that even simple strings can be converted to JS values, not just objects/arrays).  But i don't care that much :)

CyrusNajmabadi:
Thanks!

PR #302, opened 2019-07-12
Event Grid Trigger

mikhailshilkov:
Support Event Grid as Azure Functions trigger, specifically
- Blob Created/Deleted events
- Resource Group events

See the new example for API.

Covers most of #295.

### DIFF
--- a/examples/eventgrid/index.ts
+++ b/examples/eventgrid/index.ts
@@ -1,0 +1,53 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+
+const resourceGroup = new azure.core.ResourceGroup("eveentgrid-rg", {
+    location: azure.Locations.WestUS2,
+});
+
+// Subscribe to events in resource group, e.g. when a new resource is created
+resourceGroup.onGridEvent("OnResourceChange", {
+    callback: async (context, event) => {
+        context.log(`Subject: ${event.subject}`);
+        context.log(`Event Type: ${event.eventType}`);
+        context.log(`Data: ${JSON.stringify(event.data)}`);
+    },
+});
+
+// Sample storage account
+const storageAccount = new azure.storage.Account("eventgridsa", {
+    resourceGroupName: resourceGroup.name,
+    accountReplicationType: "LRS",
+    accountTier: "Standard",
+    accountKind: "StorageV2", // Only V2 accounts support Event Grid events
+});
+
+// Subscribe to creation of JPG files in any container of this storage account
+storageAccount.onGridBlobCreated("OnNewBlob", {
+    subjectFilter: {
+        caseSensitive: false,
+        subjectEndsWith: ".jpg",
+    },
+    callback: async (context, event) => {
+        context.log(`Subject: ${event.subject}`);
+        context.log(`File size: ${event.data.contentLength}`);
+    },
+});
+
+// A queue to log events to
+const logQueue = new azure.storage.Queue("log", {
+    resourceGroupName: resourceGroup.name,
+    storageAccountName: storageAccount.name,
+});
+
+// Subscribe to deletion of any files from any container of this storage account and
+// log all event data to a storage queue
+storageAccount.onGridBlobDeleted("OnDeletedBlob", {
+    outputs: [logQueue.output("log")],
+    callback: async (context, event) => {
+        context.log(`Subject: ${event.subject}`);
+        context.log(`File type: ${event.data.blobType}`);
+        return { log: event.data };
+    },
+});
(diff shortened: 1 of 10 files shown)


examples/eventgrid/index.ts
```diff
@@ -0,0 +1,53 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
```
CyrusNajmabadi: i just realized, but i don't think i've been seeing in your PRs the updates to our testing code to try to actually ensure these build and deploy.  can you update examples_tests.go to do that?  Thanks!

mikhailshilkov: Great point! Will make sure they are all there.

sdk/nodejs/eventhub/zMixins.ts
```diff
@@ -498,3 +501,180 @@ export class EventHubFunction extends appservice.Function<EventHubContext, strin
...
+            appservice.getResourceGroupNameAndLocation(args, pulumi.output(scope.resourceGroupName));
+
+        super("azure:eventhub:EventGridCallbackSubscription",
+              name,
+              new EventGridFunction(name, args),
+              { ...args, resourceGroupName, location },
+              opts);
+
+        const keys = pulumi.output(this.functionApp.getHostKeys());
+        const key = keys.apply(keys => keys.systemKeys["eventgrid_extension"]);
```
CyrusNajmabadi: should be able to write this as: `keys.systemKeys["eventgrid_extension"]` :)

sdk/nodejs/eventhub/zMixins.ts
```diff
@@ -498,3 +501,180 @@ export class EventHubFunction extends appservice.Function<EventHubContext, strin
...
+                    opts?: pulumi.ComponentResourceOptions,
+                   ): EventGridCallbackSubscription<ResourceGroupEvent>;
+    }
+}
+
+ResourceGroup.prototype.onGridEvent = function(this: ResourceGroup, name, args, opts) {
+    return new EventGridCallbackSubscription(
+        name, { id: this.id, resourceGroupName: this.name },
+        args, { parent: this, ...opts });
+};
```
CyrusNajmabadi: i'm on the fence about this.   How core do you think this functionality is?  Do people listen to resource group events a lot?  I ask since i think of ResourceGroup as a very core type, and it feels like a slightly strange one-off coupling between ResourceGroup and EventGrid.

mikhailshilkov: I have no idea, probably not too much. I've never used it before, but thought it's a decent use case. What's the alternative?

CyrusNajmabadi: as an example, for awsx we decided to invert this for things like metrics.  Instead of exposing off the resources, you can just go into each package and work with the metrics.  I somewhat feel like this should be gotten through the eventgrid module and you can say "i'd like the event grid events for these N different resource types".

mikhailshilkov: Can you give an example of a call that you have in mind? `EventGrid.onResouceGroup(resrouceGroup, {...})`?

CyrusNajmabadi: > Can you give an example of a call that you have in mind? `EventGrid.onResouceGroup(resrouceGroup, {...})?`

Right.  it could even be a top level module function.  like `eventGrid.onResourceGroupEvent`.  It might be in a sub-module.  For example, in awsx, we do this:

`awsx.cognito.metrics.compromisedCredentialsRisk(...)`.

So i'd be interested in exploring:

`azure.eventhub.onResourceGroupGridEvent(...)` or
`azure.eventhub.events.onResourceGroupdGridEvent(...)`.

sdk/nodejs/eventhub/zMixins.ts
mikhailshilkov: How does that work? I thought I can't dot into an `Output`?

CyrusNajmabadi: Yup, we added this back in 0.17: https://www.pulumi.com/blog/simplified-outputs-in-pulumi-0.17.0/

sdk/nodejs/eventhub/zMixins.ts
```diff
@@ -635,7 +635,7 @@ export class EventGridCallbackSubscription<T> extends appservice.EventSubscripti
               opts);
 
         const keys = pulumi.output(this.functionApp.getHostKeys());
-        const key = keys.apply(keys => keys.systemKeys["eventgrid_extension"]);
+        const key = keys.systemKeys["eventgrid_extension"];
```
CyrusNajmabadi: 👍

CyrusNajmabadi:
Thanks!

PR #303, opened 2019-07-13
Update to pulumi-terraform f106438be1cfc8

stack72:


stack72:
@jen20 this is an interim that fixed up something else - we need to spend more time on this app!

PR #304, opened 2019-07-15
Support input and output bindings for all trigger types

### LINKED ISSUE #232: Simplify output data flow in callback functions

mikhailshilkov:
Input and output bindings are now supported by all trigger types of callback-based Azure Functions, including single-function subscriptions and multi-function apps.

Related wiring has been moved to the base types, so each concrete function type has minimal work to do.

Added two missing samples to the test suite.

Fixes #232

### DIFF
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -205,8 +205,17 @@ interface FunctionAppArgsBase {
     readonly version?: pulumi.Input<string>;
 }
 
-export interface CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> extends CallbackArgs<C, E, R>, FunctionAppArgsBase {
-};
+/**
+ * Base arguments for all Azure Function components that are passed to MultiFunctionCallbackApp.
+ */
+export interface CallbackFunctionArgs<C extends Context<R>, E, R extends Result> extends InputOutputsArgs, CallbackArgs<C, E, R> {
+}
+
+/**
+ * Base arguments for all single-Function subscription apps.
+ */
+export interface CallbackFunctionAppArgs<C extends Context<R>, E, R extends Result> extends CallbackFunctionArgs<C, E, R>, FunctionAppArgsBase {
+}
 
 /**
  * The host.json metadata file contains global configuration options that affect all functions for a
@@ -467,12 +476,14 @@ export abstract class Function<C extends Context<R>, E, R extends Result> {
     public readonly appSettings?: pulumi.Input<{ [key: string]: string }>;
 
     constructor(name: string,
-        bindings: pulumi.Input<BindingDefinition[]>,
-        callback: CallbackArgs<C, E, R>,
-        appSettings?: pulumi.Input<{ [key: string]: string }>) {
+                trigger: pulumi.Input<InputBindingDefinition>,
+                args: CallbackFunctionArgs<C, E, R>,
+                settings?: pulumi.Input<{ [key: string]: string }>) {
+        const triggerSettings = { binding: trigger, settings: settings || {} };
+        const { bindings, appSettings } = combineBindingSettings(triggerSettings, args.inputs, args.outputs);
         this.name = name;
         this.bindings = bindings;
-        this.callback = callback;
+        this.callback = args;
         this.appSettings = appSettings;
     }
 }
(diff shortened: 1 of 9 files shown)


sdk/nodejs/appservice/zMixins.ts
```diff
@@ -773,13 +784,11 @@ export function getResourceGroupNameAndLocation(
...
-export function combineAppSettings(settings: pulumi.Input<{[key: string]: string}>[]): pulumi.Output<{[key: string]: string}> {
+function combineAppSettings(settings: pulumi.Input<{[key: string]: string}>[]): pulumi.Output<{[key: string]: string}> {
     return pulumi.all(settings).apply(items => items.reduce((a, b) => ({ ...a, ...b }), {}));
 }
 
-/** @internal */
-export function combineBindingSettings(trigger: BindingSettings<BindingDefinition>,
+function combineBindingSettings(trigger: BindingSettings<BindingDefinition>,
                                        inputs?: InputBindingSettings[],
                                        outputs?: OutputBindingSettings[]) {
```
CyrusNajmabadi: weird indentation.  can you align with "Trigger"?

CyrusNajmabadi:
LGTM.  Thanks!

PR #305, opened 2019-07-15
Fix mismatch between two merged issues

mikhailshilkov:
Two merged issues had a slight logical conflict which broke the build


CyrusNajmabadi:
👍

PR #311, opened 2019-07-23
Update azure rest versions to avoid conflict

mikhailshilkov:


stack72:
Thank you for fixing this up!! :)

PR #317, opened 2019-07-30
Update pulumi-terraform to 9b56bd2a9d

stack72:
This will allow us to be able to add the namespace aliases


stack72:
Self merging this since it has no side effects other than updating a dep - nothing in the SDK has changed

Issue #168, opened 2019-01-21
Handle migration of Azure AD resources to a separate upstream provider


jen20:
[This document](https://github.com/terraform-providers/terraform-provider-azurerm/blob/31b39db5241fb4cd3ce7805f7b692ccfcb174de1/website/docs/guides/migrating-to-azuread.html.markdown) details that Azure Active Directory management resources have been separate into their own Terraform provider and that the resources in the main provider are deprecated. We need to work out the best way to proceed there.

lukehoban:
It's possible that at this point, most of the users who would have benefitted from this have already gone through the migration process (replacing resources or manually modifying state files).  Is there still significant value in improving this?

stack72:
@lukehoban do we want to apply the work for the Aliasing that @CyrusNajmabadi did in pulumi-terraforn?

stack72:
It has been decided that the deprecation message that has been around in terraform for months has prepared our users that these are actually not maintained anymore. Adding any more messaging in Pulumi would actually confuse things

PR #321, opened 2019-07-31
Add deprecaton messages to azuread resources / datasources

stack72:
Fixes: #168

These have been migrated to https://github.com/pulumi/pulumi-azuread
This also required an upgrade to pulumi-terraform


jen20:
The deprecations look good, but do we want to standardise the English used for them - something like "please use `ServicePrincipal` from the `pulumi-azuread` package instead"?

stack72:
> The deprecations look good, but do we want to standardise the English used for them - something like "please use `ServicePrincipal` from the `pulumi-azuread` package instead"?

Works for me - you know I am always happy to take advice on messaging - if this is what we expect then it's easy to do

justinvp:
> from the `pulumi-azuread` package instead

This message will be shown for both Node and Python, right? Technically, the package names are `@pulumi/azuread` for Node and `pulumi-azuread` for Python. Should we say something that's runtime agnostic (something like "... from the Pulumi `azuread` package")? Or would it be better to list the actual package names for both Node and Python in the message (for easier `npm install`/`pip install`)?

resources.go
```diff
@@ -282,9 +282,18 @@ func Provider() tfbridge.ProviderInfo {
 		},
 		Resources: map[string]*tfbridge.ResourceInfo{
 			// Active Directory (AAD)
-			"azurerm_azuread_application":                {Tok: azureResource(azureAD, "Application")},
-			"azurerm_azuread_service_principal":          {Tok: azureResource(azureAD, "ServicePrincipal")},
-			"azurerm_azuread_service_principal_password": {Tok: azureResource(azureAD, "ServicePrincipalPassword")},
+			"azurerm_azuread_application": {
+				Tok:                azureResource(azureAD, "Application"),
+				DeprecationMessage: "please use azuread.Applicaton",
```
CyrusNajmabadi: i think we need to expand on this.  I'm not sure how a user woudl get from this message to knowing that they (presumably) need to reference a new package.  This may entail specific deprecation messages per language.  i.e. "The types and functions from the @pulumi/azure/ad module have moved to the @pulumi/azuread package."

CyrusNajmabadi:
> This message will be shown for both Node and Python, right? 

Yes.  I think the message either needs to be language specific.  Or, if it is agnostic needs to be clearer.  Like "The azure/ad module has moved to a dedicated azuread package".  The deprecation needs some indication that it's not just moving around inside this package.

If it's specific to the language, it can be much easier.  i.e. "The @pulumi/azure/ad module has moved to a dedicate @pulumi/azuread package" (and likewise for python).

stack72:
@CyrusNajmabadi / @justinvp / @jen20 so I changed this based on the feedback. The DeprecationMessage now reads as follows:

```
The azure/ad module has moved to a dedicated Pulumi azuread package
```

If we decide we need to enhance per language messaging, then we can at some point in the future

sdk/nodejs/ad/application.ts
```diff
@@ -17,6 +18,7 @@ export class Application extends pulumi.CustomResource {
      * @param state Any extra arguments used during the lookup.
      */
     public static get(name: string, id: pulumi.Input<pulumi.ID>, state?: ApplicationState, opts?: pulumi.CustomResourceOptions): Application {
+		pulumi.log.warn("Application is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
```
CyrusNajmabadi: wonky indentation.

sdk/nodejs/ad/getApplication.ts
```diff
@@ -7,7 +7,9 @@ import * as utilities from "../utilities";
 /**
  * > This content is derived from https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/d/azuread_application.html.markdown.
  */
+/** @deprecated The azure/ad module has moved to a dedicated Pulumi azuread package */
 export function getApplication(args?: GetApplicationArgs, opts?: pulumi.InvokeOptions): Promise<GetApplicationResult> & GetApplicationResult {
+	pulumi.log.warn("getApplication is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
```
CyrusNajmabadi: wonky indentation.

sdk/python/pulumi_azure/ad/application.py
```diff
@@ -8,6 +8,7 @@
 import pulumi.runtime
 from .. import utilities, tables
 
+warnings.warn("The azure/ad module has moved to a dedicated Pulumi azuread package", DeprecationWarning)
```
CyrusNajmabadi: i'm concerned about thsi.  say someone imports `azure`.  Doesn't `azure` import `ad`?  Won't that cause this path to run and warn just on import?  SHouldn't we just do this inside the constructor?

sdk/python/pulumi_azure/ad/get_application.py
```diff
@@ -66,10 +67,15 @@ def __init__(__self__, application_id=None, available_to_other_tenants=None, hom
...
     """
+    pulumi.log.warn("get_application is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
     __args__ = dict()
 
     __args__['name'] = name
     __args__['objectId'] = object_id
+    if opts is None:
+        opts = pulumi.ResourceOptions()
+    if opts.version is None:
+        opts.version = utilities.get_version()
```
CyrusNajmabadi: why did this change?

sdk/python/pulumi_azure/ad/application.py
stack72: We also have the pulumi.log.warn in the constructor:

```
pulumi.log.warn("application is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
```

The warnings.warn is just the Python-esque way of decorating like `// @deprecated...` in TypeScript

sdk/python/pulumi_azure/ad/get_application.py
stack72: This is the first time we are rolling out the version pinning code for Python - it's just a side effect of upgrading to pulumi-terraform@master

sdk/python/pulumi_azure/ad/application.py
CyrusNajmabadi: have you tested to make sure this isn't seen when someone uses python?  and only seen if they actually create the resource?

sdk/nodejs/ad/getApplication.ts
stack72: I can't see how this one is wonky... the code looks like this in the SDK

```
/**
 * > This content is derived from https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/d/azuread_application.html.markdown.
 */
/** @deprecated The azure/ad module has moved to a dedicated Pulumi azuread package */
export function getApplication(args?: GetApplicationArgs, opts?: pulumi.InvokeOptions): Promise<GetApplicationResult> & GetApplicationResult {
	pulumi.log.warn("getApplication is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
    args = args || {};
    if (!opts) {
        opts = {}
    }

    if (!opts.version) {
        opts.version = utilities.getVersion();
    }
    const promise: Promise<GetApplicationResult> = pulumi.runtime.invoke("azure:ad/getApplication:getApplication", {
        "name": args.name,
        "objectId": args.objectId,
    }, opts);

    return pulumi.utils.liftProperties(promise, opts);
}
```

is this not correct?

sdk/nodejs/ad/application.ts
stack72: Fixing this up now

sdk/nodejs/ad/getApplication.ts
CyrusNajmabadi: the indentation is off.

```typescript
	pulumi.log.warn("getApplication is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
    args = args || {};
    if (!opts) {
        opts = {}
    }
```

should be

```typescript
    pulumi.log.warn("getApplication is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package")
    args = args || {};
    if (!opts) {
        opts = {}
    }
```

CyrusNajmabadi: i think you're using tabs instead of spaces btw.

stack72: Oh I see! fixing this up now as well

sdk/python/pulumi_azure/ad/application.py
stack72: yes - I have tested both of them - you only get warnings.warn if you use the python interpretor directly

FWIW, the same things are used within classes as well

```
warnings.warn("explicit use of __name__ is deprecated", DeprecationWarning)
```

CyrusNajmabadi: > yes - I have tested both of them

Ok.  Then i'm ok :)   Just trying to be on the safe side here.

sdk/nodejs/ad/application.ts
CyrusNajmabadi: still looks wrong :D

mikhailshilkov:
I ran a `preview` with an `azure.ad.Application` component, here is the output:

```
  pulumi:pulumi:Stack (azure-play-dev):
    warning: Application is deprecated: The azure/ad module has moved to a dedicated Pulumi azuread package

  azure:ad:Application (myapp):
    warning: urn:pulumi:dev::azure-play::azure:ad/application:Application::myapp verification warning: The Azure Active Directory resources have been split out into their own Provider.

    Information on migrating to the new AzureAD Provider can be found here: https://terraform.io/docs/providers/azurerm/guides/migrating-to-azuread.html

    As such the Azure Active Directory resources within the AzureRM Provider are now deprecated and will be removed in v2.0 of the AzureRM Provider.
```

Questions:
1. Is that intended that we have two warnings for the same thing? This may be confusing, as the bigger and nicer warning points to TF instead of Pulumi.
2. Terraform's warning is actually much nicer:

    - It quotes the type and the name of the resource in question, while Pulumi warning goes under the stack.
    - It points to a URL with further explanation and action plan.
3. As already mentioned, `The azure/ad module has moved to a dedicated Pulumi azuread package` isn't very clear: I expect `azure/ad module` and `azuread package` to be confusing. Can we do a TS-specific version + a link to further guidelines?

stack72:
We are going to close this out - we don't need to worry about adding another set of messages to these resources - they have been showing as deprecated for a long time

PR #322, opened 2019-08-01
Update to v1.32.1 of the AzureRM Terraform provider

stack72:


mikhailshilkov:
Using this version on my laptop, I get 
```
creating urn:pulumi:dev::azure-table::azure:storage/account:Account::storage: Error reading 
queue properties for AzureRM Storage Account "storagef0337dd4": queues.Client#SetServiceProperties: Failure responding to request: StatusCode=403 
-- Original Error: autorest/azure: error response cannot be parsed: 
"\ufeff<?xml version=\"1.0\" encoding=\"utf-8\"?><Error><Code>AuthorizationPermissionMismatch</Code>
<Message>This request is not authorized to perform this operation using this permission.\nRequestId:ffb2702c-2003-0022-7544-48ba40000000
\nTime:2019-08-01T08:38:39.1238286Z</Message></Error>" error: invalid character 'ï' looking for beginning of value
```

which seems to be this issue https://github.com/terraform-providers/terraform-provider-azurerm/issues/3939

Not sure why it happens for me and not CI, but we should be really careful with this update.

mikhailshilkov:
As discussed with @stack72, the build fails because TF changed the format of Table IDs from `https://account1.table.core.windows.net/table1` to `https://account1.table.core.windows.net/Tables('table1')`. It seems that we need a migration, TF did it in [this PR](https://github.com/terraform-providers/terraform-provider-azurerm/pull/3932).

stack72:
I am going to close this out and reopen it - so much has changed in master that we need to do recreate this scenario

Issue #269, opened 2019-06-04
Namespace problem: azure.keyvault.Certifiate


stack72:
Typo in the namespace - should be `azure.keyvault.Certificate`

lukehoban:
Related https://github.com/pulumi/pulumi-azure/issues/97 and https://github.com/pulumi/pulumi-azure/issues/210.

stack72:
Going to close this out as it's linked to #97 

We can track all of the work in that issue rather than have multiple issues tracking a similar issue

Issue #210, opened 2019-04-17
Rename azure.eventhub.Topic to azure.servicebus.Topic


MariuszKogut:
We are using Azure ServiceBus. Is there a way to create our ServiceBus Resources using Pulumi?

lukehoban:
The servicebus resources are apparently misclassified, but are present.  They are available as `new azure.eventhub.Topic()` and other resources within the `eventhub` namespace despite not being part of `eventhub`.

We should fix the namespacing here to correctly match Azure (ideally with deprecations not hard breaks on the existing names).

mikhailshilkov:
This also applies to Event Grid which has nothing to do with Event Hubs and yet lives in its namespace. I'm doing the Event Grid triggered function and I hate to add it to `eventhubs` namespace...

Shall I try to fix this issue myself?

lukehoban:
The issue is that this will be a breaking change (as the namespace structure is part of the Pulumi identity of resources). 

This is also related to https://github.com/pulumi/pulumi-azure/issues/97 which tracks the broader breaking change we need to make to fix up namespacing. 

As noted in that issue, we have work in progress on a solution for enabling these renaming a without a breaking change. So for no I’d suggest we hold off on any changes for a bit, see if we can land that work, and then ideally make this change later when it won’t force replacements of existing resources.

stack72:
I am going to close this issue out and link it with #97 so we can track all of the migrations in that same issue

Issue #207, opened 2019-04-05
There is no azure.msi.getUserAssignedIdentity.


oliverholliday:
There is no `azure.msi.getUserAssignedIdentity`. 

I had a brief look in the terraform here: https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/azurerm/resource_arm_user_assigned_identity.go and it seems to list a 'read' method.

mikhailshilkov:
The function is located in a different namespace: `azure.core.getUserAssignedIdentity`. We should probably move it to `msi` as part of #97.

stack72:
Thanks for this @mikhailshilkov - you are indeed correct. That will be handled in this iteration

I am going to close this out and we can track all of this work in #97

Issue #97, opened 2018-09-01
Improve namespace structure 


lukehoban:
The namespacing being used is somewhat too “opinionated”, and doesn’t match the common namespace structure of Azure services as seen in SDKs. We should pick a stable reference point and cleanly align namespacing to ensure APIs are available with meaningful and correct fully qualified names.

jen20:
This would likely result in breaking changes relying on pulumi/pulumi#2389 being implemented, so I'm moving it to M22.

stack72:
There needs to be some discussion about what we are going to move and what we are not going to move. Therefore, this will not be completed in 0.25 milestone - it will move to the 0.26 milestone

PR #323, opened 2019-08-02
Restructure the Azure resources

stack72:
Fixes: #97
Fixes: #207 	
Fixes: #210 
Fixes: #269

This aligns with the same structure as that in Terraform. Test has been added
to show that the alias will work as expected

* [x] Migrate `azurerm_role` and `azurerm_user` to new `authorization` module
* [x] Migrate `azurerm_management_group` to `managementresource`
* [x] Migrate `azurerm_eventgrid__*` to new `eventgrid` module
* [x] Migrate `azurerm_servicebus_*` to new `servicebus` module
* [x] Migrate `azurerm_traffic_manager_*` to `network`
* [x] ~~Migrate `azurerm_ddos_*` to `network`~~ has been deprecated already
* [x] Rename `azure.keyvault.Certifiate` to `azure.keyvault.Certificate`
* [ ] Scan (c) banners for correct dates


lukehoban:
What about the name issue in #207, #210 and #269?

stack72:
> What about the name issue in #207, #210 and #269?

Oh, this is just the first 3 items - I thought it better to keep this PR moving a little at a time - I know that these are going as there will be a discussion regarding it

resources.go
```diff
@@ -66,12 +72,11 @@ const (
 	azureLogicApps           = "logicapps"           // Logic Apps
 	azureLB                  = "lb"                  // Load Balancer
 	azureMariaDB             = "mariadb"             // MariaDB
-	azureMessaging           = "eventhub"            // Event Hub
+	azureEventGrid           = "eventgrid"           // Event Grid
+	azureEventHub            = "eventhub"            // Event Hub
 	azureMgmtResource        = "managementresource"  // Management Resource
```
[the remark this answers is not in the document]
stack72: Do you think it should be plural? If so, now is the time to do this

examples/msi-renamed-to-authorization/package.json
```diff
@@ -0,0 +1,20 @@
+{
+    "name": "msi-renamed-to-authorization",
+    "version": "0.0.1",
+    "main": "bin/index.js",
+    "typings": "bin/index.d.ts",
+    "scripts": {
+        "build": "tsc"
+    },
+    "dependencies": {
+        "@pulumi/pulumi": "dev"
```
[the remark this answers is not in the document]
stack72: There will be a different project around this - this is following what the rest do right now

resources.go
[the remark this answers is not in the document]
mikhailshilkov: I believe most other things are plural, as in "this is a group of management resources". The "resource" name is a bit weird, maybe shorten it to "management"? Although TF does use "resource".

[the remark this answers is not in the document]
CyrusNajmabadi: We should follow terraform here.  it looks like they just call it 'management': https://www.terraform.io/docs/providers/azurerm/r/management_lock.html

resources.go
```diff
@@ -979,9 +941,8 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_cosmosdb_account":             {Tok: azureDataSource(azureCosmosDB, "getAccount")},
 			"azurerm_data_lake_store":              {Tok: azureDataSource(azureDatalake, "getStore")},
 			"azurerm_dev_test_lab":                 {Tok: azureDataSource(azureDevTest, "getLab")},
-			"azurerm_eventhub_namespace":           {Tok: azureDataSource(azureMessaging, "getEventhubNamespace")},
+			"azurerm_eventhub_namespace":           {Tok: azureDataSource(azureEventHub, "getEventhubNamespace")},
```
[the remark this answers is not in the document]
stack72: I can't do anything about case sensitivity changes :( We get TypeScript errors and every filesystem bar linux would complain

resources.go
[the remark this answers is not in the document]
stack72: done!

examples/examples_test.go
```diff
@@ -66,6 +66,26 @@ func TestExamples(t *testing.T) {
...
+			EditDirs: []integration.EditDir{
+				{
+					Dir:      "step2",
+					Additive: true,
+					ExpectNoChanges: true,
+				},
+			},
+		}),
+		base.With(integration.ProgramTestOptions{
+            Dir: path.Join(cwd, "servicebus-migration-test"),
```
mikhailshilkov: The indentation looks off? (tabs/spaces?)

examples/msi-renamed-to-authorization/.gitignore
```diff
@@ -0,0 +1,2 @@
+/bin/
+/node_modules/
```
mikhailshilkov: This file might not be required - I think both exclusions are covered in the parent? (same applies to other examples)

examples/msi-renamed-to-authorization/index.ts
```diff
@@ -0,0 +1,10 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
```
mikhailshilkov: Nit: Should this file be in `step1` folder?

examples/servicebus-migration-test/step2/index.ts
```diff
@@ -0,0 +1,28 @@
...
+const exampleNamespace = new azure.servicebus.Namespace("example", {
+    location: exampleResourceGroup.location,
+    name: "tfex_sevicebus_namespace",
+    resourceGroupName: exampleResourceGroup.name,
+    sku: "Standard",
+    tags: {
+        source: "example",
+    },
+});
+const serviceBusQueue = new azure.servicebus.Queue("example", {
```
mikhailshilkov: Should we convert all existing examples to the new namespaces? Might be a nice extra sanity check?

resources.go
```diff
@@ -550,41 +548,12 @@ func Provider() tfbridge.ProviderInfo {
...
-					// Max length of a servicehub topic is 260.
-					azureName: AutoNameWithMaxLength(azureName, 260),
-				}},
-			"azurerm_servicebus_topic_authorization_rule": {Tok: azureResource(azureMessaging, "TopicAuthorizationRule")},
+			// EventHub
+			"azurerm_eventhub":                              {Tok: azureResource(azureEventHub, "EventHub")},
+			"azurerm_eventhub_authorization_rule":           {Tok: azureResource(azureEventHub, "EventHubAuthorizationRule")},
+			"azurerm_eventhub_consumer_group":               {Tok: azureResource(azureEventHub, "EventHubConsumerGroup")},
+			"azurerm_eventhub_namespace":                    {Tok: azureResource(azureEventHub, "EventHubNamespace")},
+			"azurerm_eventhub_namespace_authorization_rule": {Tok: azureResource(azureEventHub, "EventHubNamespaceAuthorizationRule")},
```
mikhailshilkov: Now when we have a proper module name, do we still want to prefix every resource with `EventHub`?

resources.go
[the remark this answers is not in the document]
mikhailshilkov: Related to another comment, we could make it just `getNamespace`

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+	)
+	renameResourceWithAlias("azurerm_servicebus_topic_authorization_rule", "TopicAuthorizationRule", "",
+		azureEventHub, azureServiceBus, nil)
+	renameDataSourceWithAlias("azurerm_servicebus_namespace", "getServiceBusNamespace", "getNamespace",
+		azureEventHub, azureServiceBus, nil)
+
+	// Migrate `azurerm_traffic_manager_*` to network module
+	renameResourceWithAlias("azurerm_traffic_manager_endpoint", "Endpoint", "TrafficManagerEndpoint",
+		azureLegacyTrafficManager, azureNetwork, nil)
+	renameResourceWithAlias("azurerm_traffic_manager_profile", "Profile", "TrafficManagerProfile",
```
mikhailshilkov: TF seems to have `TrafficManager` in its own module. Why do we move to `network`?

examples/examples_test.go
stack72: It would fail go-lint if it was an issue :)

stack72: This is the beauty of go

examples/msi-renamed-to-authorization/.gitignore
stack72: This is copied and pasted from other examples :) We can make a pass through these later if you feel strongly about them?

examples/msi-renamed-to-authorization/index.ts
stack72: I am copying this from what @CyrusNajmabadi setup in pulumi-aws. Where we assume this is step1 and then apply a real step2 on top

resources.go
stack72: Traffic manager is bundled as part of network in the Terraform docs - this is why we went with it

@lukehoban thoughts on this one?

resources.go
stack72: Good idea! I can change these

resources.go
lukehoban: > TF seems to have TrafficManager in its own module. 

What do you mean by this?  TF has no concept of modules - right?

examples/msi-renamed-to-authorization/index.ts
```diff
@@ -0,0 +1,10 @@
...
+
+import * as azure from "@pulumi/azure";
+
+const resourceGroup = new azure.core.ResourceGroup("aks");
+
+const testIdentity = new azure.msi.UserAssignedIdentity("testIdentity", {
```
CyrusNajmabadi: test is small enough to be nbd, but it might stil be good to mention in a comment here that we're going to switch this from azure.msi to azure.authorization in step 2.

nbd though.

examples/servicebus-migration-test/index.ts
```diff
@@ -0,0 +1,25 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+
+const exampleResourceGroup = new azure.core.ResourceGroup("example");
+
+const exampleNamespace = new azure.eventhub.Namespace("example", {
```
CyrusNajmabadi: i'd def comment here since there are multiple things you're moving

examples/servicebus-migration-test/step2/index.ts
CyrusNajmabadi: yes, we should.  though that can come in another PR imo.  but it woudl just be good so anyone looking at it is pushed to the right path.

resources.go
```diff
@@ -83,16 +88,22 @@ const (
...
 	azureStorage             = "storage"             // Storage
 	azureStreamAnalytics     = "streamanalytics"     // StreamAnalytics
-	azureTrafficManager      = "trafficmanager"      // Traffic Manager
+
+	// Legacy Module Names
+	azureLegacyRole             = "role"               // Azure Role
+	azureLegacyMSI              = "msi"                // Managed Service Identity (MSI)
+	azureLegacyManagementGroups = "managementgroups"   // Management Groups (Legacy)
+	azureLegacyMgmtResource     = "managementresource" // Management Resource (Legacy)
+	azureLegacyTrafficManager   = "trafficmanager"     // Traffic Manager
```
CyrusNajmabadi: can you be consistent here with the `(legacy)` moniker?  Having some have it and some not actually confuses me in terms of wondering if they're different somehow.

resources.go
CyrusNajmabadi: agreed.  Can we strip this before going in.

resources.go
```diff
@@ -606,7 +575,6 @@ func Provider() tfbridge.ProviderInfo {
 				},
 			},
 			"azurerm_key_vault_access_policy": {Tok: azureResource(azureKeyVault, "AccessPolicy")},
-			"azurerm_key_vault_certificate":   {Tok: azureResource(azureKeyVault, "Certifiate")},
```
CyrusNajmabadi: ah... fixing mispelling.  oops.

resources.go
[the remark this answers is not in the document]
CyrusNajmabadi: Yes.  please just make getNamespace.  Now it avoids all the problems! :)

resources.go
```diff
@@ -1046,11 +990,10 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_application_security_group": {Tok: azureDataSource(azureNetwork, "getApplicationSecurityGroup")},
 			"azurerm_recovery_services_vault":    {Tok: azureDataSource(azureRecoveryServices, "getVault")},
 			"azurerm_redis_cache":                {Tok: azureDataSource(azureRedis, "getCache")},
-
-			"azurerm_resource_group": {Tok: azureDataSource(azureCore, "getResourceGroup")},
-			"azurerm_snapshot":       {Tok: azureDataSource(azureCompute, "getSnapshot")},
-			"azurerm_subnet":         {Tok: azureDataSource(azureNetwork, "getSubnet")},
-			"azurerm_route_table":    {Tok: azureDataSource(azureNetwork, "getRouteTable")},
+			"azurerm_resource_group":             {Tok: azureDataSource(azureCore, "getResourceGroup")},
+			"azurerm_snapshot":                   {Tok: azureDataSource(azureCompute, "getSnapshot")},
```
CyrusNajmabadi: aside, do we think azureCompute is the right namespace for this resource?  i honestly don't know.  but 'snapshot' sounds odd there.

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+		}
+
+		currentInfo.Tok = azureResource(currentModule, tokName)
+		currentInfo.Aliases = []tfbridge.AliasInfo{
+			{Type: &legacyType},
+		}
+
+		if legacyInfo.Docs == nil {
+			legacyInfo.Docs = &tfbridge.DocInfo{
+				Source: tfName[len("azurerm_"):] + ".html.markdown",
```
CyrusNajmabadi: just to check, but have you validated this works?  This was right for teh aws provider, but really depends on what terraform does for their docs.

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+			}
+		}
+
+		prov.DataSources[tfName] = &currentInfo
+		prov.DataSources[legacyTFName] = &legacyInfo
+		prov.P.DataSourcesMap[legacyTFName] = prov.P.DataSourcesMap[tfName]
```
CyrusNajmabadi: didn't look closely at thsi :)   i assumed it was copied mostly verbatim from what i did in AWS?  If that's **not** the case, and you made any interesting changes, can you please let me know?

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+
+	// New Authorization Mod - this combines the old MSI and Role Modules
+	renameResourceWithAlias("azurerm_role_assignment", "Assignment", "", azureLegacyRole, azureAuthorization,
+		&tfbridge.ResourceInfo{
+			Fields: map[string]*tfbridge.SchemaInfo{
+				// Suppress auto-naming of this field. It is autonamed to a GUID in the underlying provider.
+				azureName: {Name: azureName},
+			},
+		},
+	)
```
CyrusNajmabadi: it's effectively hard to review this bit (becaues of all the jumping up and down).  To the best of your knowledge have you gone and double-checked that you copy/pasted this stuff correctly, and have you at least looked at the final modules to make sure they match your expectations?

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+	renameResourceWithAlias("azurerm_role_definition", "Definition", "", azureLegacyRole, azureAuthorization, nil)
```
CyrusNajmabadi: why is this `azure.authorization.Definition` and not `azure.authorization.RoleDefinition`?

resources.go
```diff
@@ -1293,3 +1244,192 @@ func FromName(options AutoNameOptions) func(res *tfbridge.PulumiResource) (inter
...
+	renameResourceWithAlias("azurerm_user_assigned_identity", "UserAssignedIdentity", "", azureLegacyMSI,
+		azureAuthorization, &tfbridge.ResourceInfo{
+			Docs: &tfbridge.DocInfo{
+				Source: "user_assigned_identity.markdown",
+			},
+		},
+	)
+	renameDataSourceWithAlias("azurerm_role_definition", "getRoleDefinition", "",
```
CyrusNajmabadi: this is def confusing since it's getRoleDefinition here, but Definition above.

resources.go
CyrusNajmabadi: Yup, given that it's under here:

image

I'm ok with this being in the 'network' module for us.

CyrusNajmabadi:
1. It's a pita, but please also delete the previous files that were generated that are no longer being generated.  
2. we def dont' want to have to fix any of these up again.  So it would be great to just go over this one more time with a fine-toothed comb making sure that the names match expected patterns, and the right stuff was copied from the before/after.  

There were a few tiny discrepancies that make me think we have some issues.  Basically, the overall idea is sound, but we have to nail the execution here so we don't end with additional mess we have to clean up in the near term.  Thanks! :)

resources.go
CyrusNajmabadi: Resolved.  Checked tf and this is where it lives.

resources.go
mikhailshilkov: Yeah, I guess we can ignore that. I just wasn't sure Traffic Manager belongs to network, but it might.

resources.go
stack72: :)

resources.go
stack72: done!

resources.go
stack72: Yes - I am still getting the docs on the providers as expected :)

resources.go
stack72: yes - I made a change to it that allows renaming of the Tok that's the only change I made

stack72: FWIW, this is why the tests were added early - to validate :)

resources.go
stack72: I passed the ability to use a new Tok in the new Namespace and then we use the old one in the old Namespace for backwards compat

resources.go
stack72: done!

resources.go
stack72: done

resources.go
stack72: done

stack72:
@CyrusNajmabadi / @mikhailshilkov ok, so I took a final pass. Unfortunately, there are a few things in the Eventhub namespace that can't strip Eventhub from the Tok names as EventGrid and ServiceBus names collided. I removed as many as I could

I believe this is ready for a final review!

mikhailshilkov:
> Unfortunately, there are a few things in the Eventhub namespace that can't strip Eventhub from the Tok names as EventGrid and ServiceBus names collided. I removed as many as I could

I vote for deprecating the whole `eventhub` module and moving the Azure Event Hubs resources to a new `eventhubs` module. It feels better than having `eventhub.Namespace` (deprecated Service Bus)  and `eventhub.EventHubNamespace` (actual Event Hubs) next to each other. @stack72 @CyrusNajmabadi

CyrusNajmabadi:
@mikhailshilkov @stack72 discussed it with @lukehoban . The value in having the names be totally crisp doesn't outweigh the negative of now having an `eventhub` *and* `eventhubs` namespace we need to maintain.  

So, for now, we should do:

1. the current crop of renames.
2. have `eventhub.EventHubNamespace`
3. deprecate the older resources.  

In teh short term, we should ensure that we generate the `@deprecation` tags and that this properly removes the resources from docs.

In the medium/long term we would like to deeply follow through on deprecation, including 
1. starting to warn people to not use these resources (and tell them which to use instead).
2. remove these resources.  because we use aliases, it should be possible for people to move without breaks.

At the point that we have gone through all that, we can *consider* then renaming `EventHubNamespace` to `Namespace`.

examples/servicebus-migration-test/step2/index.ts
```diff
@@ -0,0 +1,28 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+
+const exampleResourceGroup = new azure.core.ResourceGroup("example", {
+    location: "West Europe",
+    name: "my-servicebus",
+});
+const exampleNamespace = new azure.servicebus.Namespace("example", {
+    location: exampleResourceGroup.location,
```
mikhailshilkov: The preference is to avoid assigning locations explicitly and do so from the config.

resources.go
mikhailshilkov: This seems to produce broken links for legacy resources, e.g. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/eventgrid_topic_legacy.html.markdown

CyrusNajmabadi:
> This seems to produce broken links for legacy resources, e.g. https://github.com/terraform-providers/terraform-provider-azurerm/blob/master/website/docs/r/eventgrid_topic_legacy.html.markdown

Not great, but probably not the end of hte world.  Should we just have the legacy resource pointing at the docs for the modern resource?  it would be nice to fix this, but i wouldn't block on this

PR #325, opened 2019-08-03
Remove explicit location assignments from examples

mikhailshilkov:
We want to steer users away from hard-coding locations in their programs, so we remove those from our own examples


mikhailshilkov:
@lukehoban done

PR #328, opened 2019-08-07
Fix location propagation of callback functions

mikhailshilkov:
Apparently, we don't respect a `location` argument passed to callback functions. E.g.

```ts
const resourceGroup = new azure.core.ResourceGroup("UrlShorterner", {
    location: "westeurope",
});

const fn = new azure.appservice.HttpEventSubscription("test", {
    resourceGroup,
    location: "westus",
    callback,
});
```

was placed in West Europe, not West US. The PR fixes that.


sdk/nodejs/appservice/zMixins.ts
```diff
@@ -772,14 +772,18 @@ export function getResourceGroupNameAndLocation(
...
     }
 
     if (!args.resourceGroupName && !fallbackResourceGroupName) {
         throw new Error("Either [args.resourceGroup] or [args.resourceGroupName] must be provided.");
     }
 
     const resourceGroupName = util.ifUndefined(args.resourceGroupName, fallbackResourceGroupName!);
+    if (args.location) {
+        return { resourceGroupName, location: args.location };
+    }
```
CyrusNajmabadi: might be cleaner as:

```ts
const location = args.location || resourceGroupName.apply(n => core.getResourceGroup({ name: n })).location;
return { resourceGroupName, location };
```

mikhailshilkov: Love it!

PR #337, opened 2019-08-22
Upgrade to v1.33.0 of the Azurerm Terraform Provider

stack72:


stack72:
Just an FYI with this, I believe 3 of the resources are incorrectly namedx in the upstream provider - I have questioned this with the Terraform team but right now, we are mapped correctly according to their system

jen20:
This is superseded by #337 subject to the same constraints and notes - that updates us to 1.33.1 instead of 1.33.0.

PR #341, opened 2019-08-27
Update to v1.33.1 of the Terraform AzureRM provider

jen20:
This updates us to the latest patch release of the Terraform AzureRM provider. Since #337 is not yet merged, this PR supersedes that since it maps all new resources that were not on `master` also.

It is still subject to the same release constraints as the work in #337.


stack72:
Closed via #344 - rebase was more difficult after more changes to pulumi-terraform!

PR #342, opened 2019-08-28
Add the Get Started button and prompt

bermudezmt:
I had the text match what we have in our Get Started callout on the docs site.


stack72:
@bermudezmt unfortunately these badges are not correct - they should not link to @pulumi/pulumi - they should be specific to pulumi-azure

I will send a PR to fix these

PR #354, opened 2019-09-05
Add a wait of Azure Function availability 

### LINKED ISSUE #353: EventGridCallbackSubscription sometimes fails (which fails the build)
Some builds ([example](https://travis-ci.com/pulumi/pulumi-azure/builds/125850614#L4100)) fail with the following error:

```
Plan apply failed: Error waiting for EventGrid Event Subscription "ondeletedblob2442c348" 
(Scope "/subscriptions/0282681f-7a9e-424b-80b2-96babd57a8a1/resourceGroups/eveentgrid-rg93db5886/providers/Microsoft.Storage/storageAccounts/eventgridsa86ade982") 
to become available: Code="Url validation" Message="The attempt to validate the provided
endpoint https://ondeletedblobc7d743dd.azurewebsites.net/runtime/webhooks/eventgrid failed.
For more details, visit https://aka.ms/esvalidation."
```

This comes from our EventGrid mix-ins where [we set the URL of an Event Grid subscription to point to the Function App](https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/eventgrid/zMixins.ts#L134-L140).

There is no exact HTTP error in the log that the subscription gets. Most probably, the Function App isn't ready yet so it returns some sort of error which Event Grid interprets as failed validation.

A deep fix would involve either a change in Function Apps, or retries built into the subscription provisioning. I don't think this can be easily done in Terraform, probably it's somewhere further upstream.

We could probably add a validation of the Function App endpoint in our own code. Is there a way to wrap the [`url`](https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/eventgrid/zMixins.ts#L134) output into another output which would wait until this URL returns a `200`?

mikhailshilkov:
Add a wait of Azure Function availability before linking an event grid subscription to it. The code will send a sample request, check if the response is 2xx, and retry otherwise for 5 minutes max until it is.

Resolves #353

### DIFF
--- a/sdk/nodejs/eventgrid/zMixins.ts
+++ b/sdk/nodejs/eventgrid/zMixins.ts
@@ -15,6 +15,7 @@
 import * as pulumi from "@pulumi/pulumi";
 
 import * as eventgrid from "azure-eventgrid/lib/models";
+import fetch from "node-fetch";
 import * as appservice from "../appservice";
 import { ResourceGroup } from "../core";
 import { Account } from "../storage";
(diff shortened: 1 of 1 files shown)


stack72:
@mikhailshilkov this is awesome :)

mikhailshilkov:
@stack72 One thing I'm thinking - maybe do no throw an Error after the timeout and just let the Event Grid make a final try? In case they change something and our fake request becomes incompatible. WDYT?

stack72:
@mikhailshilkov let's try this first! If this doesn't feel nice for us before the next release of the provider, then we can enhance it

sdk/nodejs/eventgrid/zMixins.ts
```diff
@@ -132,9 +133,10 @@ export class EventGridCallbackSubscription<T> extends appservice.EventSubscripti
         const keys = pulumi.output(this.functionApp.getHostKeys());
         const key = keys.systemKeys["eventgrid_extension"];
         const url = pulumi.interpolate`https://${this.functionApp.defaultHostname}/runtime/webhooks/eventgrid?functionName=${name}&code=${key}`;
+        const liveUrl = url.apply(u => waitUntilEndpointIsUp(this, u));
```
lukehoban: Is it guaranteed that this url endpoint is internet addressable?

sdk/nodejs/eventgrid/zMixins.ts
```diff
@@ -143,6 +145,35 @@ export class EventGridCallbackSubscription<T> extends appservice.EventSubscripti
...
+            status = `${response.status}: ${response.statusText}`;
+        } catch (e) {
+            status = `Error: ${e}`;
+        }
+        pulumi.log.info(`Waiting for Webhook ${url} to become available (${status}, ${i})`, resource);
+        // Wait for 10s between polls
+        await new Promise(r => setTimeout(r, 10000));
+    }
+
+    throw new Error("timed out waiting for Webhook to become up");
```
lukehoban: Is "Webhook" the right word for this?

sdk/nodejs/eventgrid/zMixins.ts
mikhailshilkov: As of today, yes, Event Grid requires publicly accessible endpoints. Private endpoints require tricks like putting a proxy or a hybrid connection in between.

sdk/nodejs/eventgrid/zMixins.ts
mikhailshilkov: We wait for `https://app.azurewebsites.net/runtime/webhooks/eventgrid` so I called it webhook everywhere. I'm okay to change the wording if you think it can make things more clear?

stack72:
I’m happy for this to be merged - we can let the examples run through and test it :) thanks so much @mikhailshilkov

Issue #356, opened 2019-09-06
package fails to install on windows


4c74356b41:
```
> pip3 install pulumi-azure                                                                                             Collecting pulumi-azure
  Using cached https://files.pythonhosted.org/packages/8f/76/1cf9e1e4e1e8d13222a34034d388fcb03d4d973f46cc1249e4caa9bb6bbf/pulumi_azure-1.0.0.tar.gz
    ERROR: Command errored out with exit status 1:
     command: 'c:\users\core\appdata\local\programs\python\python36-32\python.exe' -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'C:\\Users\\core\\AppData\\Local\\Temp\\pip-install-n1bhdol8\\pulumi-azure\\setup.py'"'"'; __file__='"'"'C:\\Users\\core\\AppData\\Local\\Temp\\pip-install-n1bhdol8\\pulumi-azure\\setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' egg_info --egg-base pip-egg-info
         cwd: C:\Users\core\AppData\Local\Temp\pip-install-n1bhdol8\pulumi-azure\
    Complete output (9 lines):
    Traceback (most recent call last):
      File "<string>", line 1, in <module>
      File "C:\Users\core\AppData\Local\Temp\pip-install-n1bhdol8\pulumi-azure\setup.py", line 33, in <module>
        long_description=readme(),
      File "C:\Users\core\AppData\Local\Temp\pip-install-n1bhdol8\pulumi-azure\setup.py", line 28, in readme
        return f.read()
      File "c:\users\core\appdata\local\programs\python\python36-32\lib\encodings\cp1252.py", line 23, in decode
        return codecs.charmap_decode(input,self.errors,decoding_table)[0]
    UnicodeDecodeError: 'charmap' codec can't decode byte 0x9d in position 452: character maps to <undefined>
    ----------------------------------------
ERROR: Command errored out with exit status 1: python setup.py egg_info Check the logs for full command output.
```

windows 10, pulumi 1.0, python 3.6.8. venv\no venv doesnt make a difference. works fine on linux

ellismg:
[...]

PR #357, opened 2019-09-06
Fix package install on Windows

ellismg:
Our README.rst (which is generated from README.md) has smart quotes in
it, which can break install of the package on Windows (since reading
the README.rst fails because it is not valid in the current windows
code-page).

This change ensures we don't generate such `.rst` files. This is the
quickest thing to to address the issue - a longer term fix I'll land
everywhere soon is to just always open these files as UTF-8, and
hopefully remove the need to use `.rst` files at all.

Fixes: #356

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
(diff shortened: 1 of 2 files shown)


ellismg:
Possible (depending on their `README.md`'s.)  My plan is to fix up CI so we use Python 3 everywhere today, which will let me land https://github.com/pulumi/pulumi-terraform/pull/388 which will force UTF-8 encoding when opening these files.

PR #361, opened 2019-09-10
Fix the usage of NsRecord.records

### LINKED ISSUE #243: Errors creating NsRecord.
When trying to create an NsRecord entry for a DNS Zone in an existing Zone I hit some snags. 

```typescript
// set up
const rootZone = await azure.dns.getZone({
    name: "dev1.test.com",
    resourceGroupName: "Infrastructure"
});

const dnsZone = new azure.dns.Zone("z1-dev1-dns-zone", {
    name: "z1.dev1.test.com",
    resourceGroupName: rootZone.resourceGroupName,
});
```

The following (and variations) didn't work. Note that I'm using the `records` field not `record` yet I receive a verification error for `record`.

The type definition for `NsRecordArgs.records` is this, but it doesn't seem to be correct (object expected):
```typescript
readonly records?: pulumi.Input<pulumi.Input<string>[]>;
```

```typescript
// this failed with errors and a deprecation warning:
//    verification warning: "record": [DEPRECATED] This field has been replaced by `records`
//    error [snip] has a problem: record.0: expected object, got string
//    error [snip] has a problem: record.1: expected object, got string
//    error [snip] has a problem: record.2: expected object, got string
//    error [snip] has a problem: record.3: expected object, got string
new azure.dns.NsRecord("z1-dev1-ns-record", {
    name: "z1",
    resourceGroupName: rootZone.resourceGroupName,
    zoneName: rootZone.name,
    records: dnsZone.nameServers,
    ttl: 3600
});
```

This did work, but it is cumbersome and marked as deprecated:

```typescript
// this did work but get the deprecation warning : 
// verification warning: "record": [DEPRECATED] This field has been replaced by `records`
new azure.dns.NsRecord("z1-dev1-ns-record", {
    name: "z1",
    resourceGroupName: rootZone.resourceGroupName,
    zoneName: rootZone.name,
    record: pulumi.output(dnsZone.nameServers).apply(x => x.map(y => ({ nsdname: y }))),
    ttl: 3600
});

```

mikhailshilkov:
Fixes #243 
Technically, this seems a breaking change. Somebody could have used `records` to assign `record`-compatible array and that would probably work. Hopefully, that's marginal enough to actually include the fix into the next version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,8 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Fix an issue where `pulumi-azure` would fail to install on Windows (fixes [#356](https://github.com/pulumi/pulumi-azure/issues/356))
+* Fix an issue where `records` property of `dns.NsRecord` was mistakenly mapped to the deprecated `record` property (fixes [#243](https://github.com/pulumi/pulumi-azure/issues/243))
+  [#361](https://github.com/pulumi/pulumi-azure/pull/361)
 
 ---
 
--- a/resources.go
+++ b/resources.go
@@ -539,9 +539,12 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_dns_ns_record": {
 				Tok: azureResource(azureDNS, "NsRecord"),
 				Fields: map[string]*tfbridge.SchemaInfo{
-					// This property is deprecated and renamed to `records`.  Don't pluralize this depreacted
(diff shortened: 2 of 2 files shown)


stack72:
So this is a good fix - the usecase that *could* have worked isn’t documented anywhere and isn’t what our examples or docs suggest. So it’s a coincidence the user found it. I’m going to suggest this LGTM and we will be able to release this in a minor version bump

It’s not breaking as it won’t cause the recreation of any resources

PR #378, opened 2019-09-30
Use Blob SAS instead of Account SAS for better security

mikhailshilkov:
The latest version of Terraform introduced a data source for Blob Container-level Shared Access Signature. It has a smaller scope than the Account-level SAS, so let's use it for our zip files for function packages.


stack72:
@mikhailshilkov this LGTM! I just manually merged this to get rid of the merge conflict in the CHANGELOG - https://github.com/pulumi/pulumi-azure/commit/0c39f4546070e26dc61f7b7e6d8d94b16ef3792b

PR #414, opened 2019-11-19
Fix broken link on userAssignedIdentity docs

kenny-wealth:
- Fixing a broken link on authorization docs
- Add /obj to .gitignore

### DIFF
--- a/sdk/dotnet/.gitignore
+++ b/sdk/dotnet/.gitignore
@@ -1,1 +1,2 @@
 version.txt
+obj/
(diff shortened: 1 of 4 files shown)


pgavlin:
Thanks for the PR, @kenny-wealth ! These links are auto-generated by code in https://github.com/pulumi/pulumi-terraform-bridge, so unfortunately these fixes will be overwritten the next time the code is regenerated. @stack72 can you help figure out where in tfgen we need to make a fix?

kenny-wealth:
> Thanks for the PR, @kenny-wealth ! These links are auto-generated by code in https://github.com/pulumi/pulumi-terraform-bridge, so unfortunately these fixes will be overwritten the next time the code is regenerated. @stack72 can you help figure out where in tfgen we need to make a fix?

Cool. I'll just go ahead and close the PR.

stack72:
Thanks @kenny-wealth 

I’m actually going to get this fixed and rolled out today as I’ve found the cause of the problem 

P.

PR #419, opened 2019-11-26
Upgrade to v1.37.0 of the AzureRM Terraform Provider

stack72:


stack72:
@mikhailshilkov ill look at these on this flight - I’m not sure why these are happening in CI and not locally :/

stack72:
PR fails due to serviceBus not found which was fixed outside of this - merging so we can get a full test run in master

PR #422, opened 2019-11-29
Add a test for a C# example

mikhailshilkov:


mikhailshilkov:
@lukehoban I also changed the build scripts to copy the nuget package to the local source

PR #425, opened 2019-12-03
Add explicit names for all C# namespaces

mikhailshilkov:
A follow-up on https://github.com/pulumi/pulumi-terraform/pull/510 and a step towards https://github.com/pulumi/pulumi-terraform/issues/511
This adds explicit names for all C# namespaces in Azure.

Further plan:
1. Approve this PR.
2. Do a similar change for all providers.
3. Change TF Gen to remove the fallback to `Titlecase` and require the namespace to be explicitly provided.
4. Promote `Namespaces` property from `CSharpInfo` to top-level `Modules`, allowing any new language generator to rely on it.
5. Optionally, retrofit this definition to old language gens to make ToLower on a module name instead of using `Tok` directly.

### DIFF
--- a/resources.go
+++ b/resources.go
@@ -44,79 +44,83 @@ const (
 	azurePkg = "azure"
 	// modules; in general, we took naming inspiration from the Azure SDK for Go:
 	// https://godoc.org/github.com/Azure/azure-sdk-for-go
-	azureAD                  = "ad"                  // Active Directory (AAD)
-	azureAnalysisServices    = "analysisservices"    // Analysis Services
-	azureAPIManagement       = "apimanagement"       // API Management
-	azureAppInsights         = "appinsights"         // AppInsights
-	azureAppService          = "appservice"          // App Service
-	azureAutomation          = "automation"          // Automatio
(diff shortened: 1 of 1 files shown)


resources.go
```diff
@@ -128,18 +132,29 @@ func azureType(mod string, typ string) tokens.Type {
 	return tokens.Type(azureMember(mod, typ))
 }
 
+// azureSubType manufactures a type token for the Azure package and the given module, submodule, and type.
+func azureSubType(moduleTitle, sub, typ string) tokens.Type {
+	moduleName := strings.ToLower(moduleTitle)
+	namespaceMap[moduleName] = moduleTitle
+	return azureType(moduleName+"/"+sub, typ)
```
pgavlin: How about refactoring s.t. `azure{Type,Member}` take the module name and type/function name separately? Then we could just do `strings.ToLower` in one/both of those functions and let it do the `module+"/"+thing` dance.

mikhailshilkov: Good call! I moved all the logic to `azureMember` and other functions are just one-liners with descriptive names.

PR #427, opened 2019-12-06
Upgrade to v1.5.0 of the pulumi-terraform-bridge

stack72:


mikhailshilkov:
The change LGTM but it looks like we need another release of Pulumi to make this work https://github.com/pulumi/pulumi/pull/3614

stack72:
Closing until we can get pulumi/pulumi#3614 merged

PR #436, opened 2019-12-16
Upgrade to v1.39.0 of the AzureRM Terraform Provider

stack72:


resources.go
```diff
@@ -47,6 +47,7 @@ const (
 	azureAD                  = "AD"                  // Active Directory (AAD)
 	azureAnalysisServices    = "AnalysisServices"    // Analysis Services
 	azureAPIManagement       = "ApiManagement"       // API Management
+	azureAppConfiguration    = "AppConfiguration"    // AppConfiguration
```
mikhailshilkov: Nit to match the product name
```suggestion
	azureAppConfiguration    = "AppConfiguration"    // App Configuration
```

resources.go
```diff
@@ -838,6 +839,15 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_mysql_firewall_rule":        {Tok: azureResource(azureMySQL, "FirewallRule")},
 			"azurerm_mysql_server":               {Tok: azureResource(azureMySQL, "Server")},
 			"azurerm_mysql_virtual_network_rule": {Tok: azureResource(azureMySQL, "VirtualNetworkRule")},
+			"azurerm_mssql_database_vulnerability_assessment_rule_baseline": {
+				Tok: azureResource(azureMySQL, "DatabaseVulnerabilityAssessmentRuleBaseline"),
```
mikhailshilkov: This is not MySQL, this is MS SQL (same for resources below)

resources.go
```diff
@@ -1194,6 +1204,7 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_bot_channel_email":         {Tok: azureResource(azureBot, "ChannelEmail")},
 			"azurerm_bot_channel_slack":         {Tok: azureResource(azureBot, "ChannelSlack")},
 			"azurerm_bot_web_app":               {Tok: azureResource(azureBot, "WebApp")},
+			"azurerm_bot_channel_ms_teams":      {Tok: azureResource(azureBot, "ChannelMsTeams")},
```
mikhailshilkov: Shouldn't `MS` be capital? At least, it should in C# naming conventions. Alternatively, I'm fine with the name "ChannelTeams".

resources.go
```diff
@@ -1210,6 +1221,9 @@ func Provider() tfbridge.ProviderInfo {
 			// NetApp
 			"azurerm_netapp_account": {Tok: azureResource(azureNetapp, "Account")},
 			"azurerm_netapp_pool":    {Tok: azureResource(azureNetapp, "Pool")},
+
+			//AppConfiguration
+			"azurerm_app_configuration": {Tok: azureResource(azureAppConfiguration, "ConfigurationStore")},
```
mikhailshilkov: I'm happy with `ConfigurationStore`. `Store` could read more natural in TypeScript, but is more explicit in namespace-less .NET code.

resources.go
stack72: I Like `ChannelTeams` better :) Moving to that!

resources.go
stack72: 👍

resources.go
stack72: Great spot!!!

PR #437, opened 2020-01-08
Upgrade to v1.40.0 of the AzureRM Terraform Provider

stack72:


resources.go
```diff
@@ -1219,11 +1234,33 @@ func Provider() tfbridge.ProviderInfo {
...
+			"azurerm_netapp_volume":   {Tok: azureResource(azureNetapp, "Volume")},
+			"azurerm_netapp_snapshot": {Tok: azureResource(azureNetapp, "Snapshot")},
 
 			//AppConfiguration
 			"azurerm_app_configuration": {Tok: azureResource(azureAppConfiguration, "ConfigurationStore")},
+
+			// Backup
+			"azurerm_backup_container_storage_account": {Tok: azureResource(azureBackup, "ContainerStorageAccount")},
+			"azurerm_backup_policy_file_share":         {Tok: azureResource(azureBackup, "PolicyFileshare")},
+			"azurerm_backup_protected_file_share":      {Tok: azureResource(azureBackup, "ProtectedFileShare")},
```
mikhailshilkov: The two resources above have different capitalization: `Fileshare` vs `FileShare`. Should be `FileShare` I believe, since Azure spell them as two words.

resources.go
```diff
@@ -1219,11 +1234,33 @@ func Provider() tfbridge.ProviderInfo {
...
+			"azurerm_backup_policy_vm":                 {Tok: azureResource(azureBackup, "PolicyVm")},
+			"azurerm_backup_protected_vm":              {Tok: azureResource(azureBackup, "ProtectedVm")},
```
mikhailshilkov: I think "Vm" should be "VM" everywhere (let me know if we already had examples of otherwise)

resources.go
stack72: good catch! Fixed

resources.go
stack72: fixed

PR #439, opened 2020-01-16
Upgrade to v1.41.1 of the AzureRM Terraform Provider

stack72:


resources.go
```diff
@@ -611,6 +614,8 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_shared_image_gallery": {Tok: azureResource(azureCompute, "SharedImageGallery")},
 			"azurerm_shared_image_version": {Tok: azureResource(azureCompute, "SharedImageVersion")},
 			"azurerm_bastion_host":         {Tok: azureResource(azureCompute, "BastionHost")},
+			"azurerm_dedicated_host_group": {Tok: azureResource(azureCompute, "DedicatedHostGroup")},
+			"azurerm_disk_encryption_set":  {Tok: azureResource(azureCompute, "DiskEncryptionSet")},
```
mikhailshilkov: Dumb question: I think we used to sort resources within each group alphabetically, but now they are kind of messed up. Do we have a convention about that?

resources.go
```diff
@@ -1424,8 +1430,14 @@ func Provider() tfbridge.ProviderInfo {
...
-			"azurerm_virtual_hub":     {Tok: azureDataSource(azureNetwork, "getVirtualHub")},
-			"azurerm_signalr_service": {Tok: azureDataSource(azureSignalr, "getService")},
+			"azurerm_virtual_hub":                 {Tok: azureDataSource(azureNetwork, "getVirtualHub")},
+			"azurerm_signalr_service":             {Tok: azureDataSource(azureSignalr, "getService")},
+			"azurerm_storage_container":           {Tok: azureDataSource(azureStorage, "getStorageContainer")},
+			"azurerm_iothub_shared_access_policy": {Tok: azureDataSource(azureIot, "getSharedAccessPolicy")},
+			"azurerm_iothub_dps":                  {Tok: azureDataSource(azureIot, "getDps")},
+			"azurerm_eventgrid_topic":             {Tok: azureDataSource(azureEventGrid, "getTopic")},
+			"azurerm_disk_encryption_set":         {Tok: azureDataSource(azureEventGrid, "getDiskEncryptionSet")},
+			"azurerm_dedicated_host_group":        {Tok: azureDataSource(azureEventGrid, "getDedicatedHostGroup")},
```
mikhailshilkov: These last two should be `compute` if I read it right

resources.go
stack72: There isn't any convention around this - we can do it but there are so many resources that it's hard to keep up

resources.go
stack72: great catch!

PR #464, opened 2020-02-25
Add support for durable functions

ChristianEder:
This PR adds support for creating Azure Durable Functions more easily - see example added.

I remember that there was a 'examples_test.go' file where I'd like to add my example to - but since my last PR this changed quite a bit and I'm not sure what is required there.


mikhailshilkov:
@ChristianEder Could you please rebase on master (it's v2.0 now) and add a test that would be a copy of https://github.com/pulumi/pulumi-azure/blob/master/examples/examples_test.go#L84 but for your new example?

ChristianEder:
@mikhailshilkov done.

ChristianEder:
hmm the build seems to fail, but I can't really figure out why - the only error it states is that there are changes in the working tree - in the files that existed before my PR, which i changed and another one that got somehow added by the build:

- changed: `sdk/nodejs/appservice/index.ts`
- added by build: `packages-microsoft-prod.deb`

any ideas?

mikhailshilkov:
@ChristianEder Yes, you need to add your new file to the list here: https://github.com/pulumi/pulumi-azure/blob/master/resources.go#L1416

ChristianEder:
@mikhailshilkov hmm i added the new mixin file to the `resources.go` file, but the build still fails, now with changes in the ` sdk/nodejs/tsconfig.json` file (which I didnt change in my PR), and still the `packages-microsoft-prod.deb` file.

mikhailshilkov:
Yeah, now you need to change `tsconfig.json` and commit it. It would be done by the tooling if you could run `make` but I guess you are on Windows and then it's non-trivial.

ChristianEder:
wohoo, green! Thanks @mikhailshilkov for your help, I totally forgot about all these places that I need to add new mixins to...

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+
+/**
+ * Azure Durable Activity Function
+ */
+export class DurableActivityFunction<TActivityInputBinding> extends mod.Function<DurableActivityFunctionContext<TActivityInputBinding>, void, void> {
+    constructor(name: string, args: DurableActivityFunctionArgs<TActivityInputBinding>) {
+        const trigger = {
+            name: args.activityInputName,
+            type: "activityTrigger",
+            direction: "in"
```
mikhailshilkov, line 63:
```suggestion
            direction: "in",
```

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+        } as mod.InputBindingDefinition;
+
+        super(name, trigger, args);
+    }
+}
+
+export interface DurableEntityFunctionContext extends mod.Context<mod.Result> {
+    df: any
```
mikhailshilkov, line 71:
```suggestion
    df: any;
```

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+export interface DurableEntityFunctionArgs extends mod.CallbackFunctionArgs<DurableEntityFunctionContext, void, void> {
+}
+
+/**
+ * Azure Durable Entity Function
+ */
+export class DurableEntityFunction extends mod.Function<DurableEntityFunctionContext, void, void> {
+    constructor(name: string, args: DurableEntityFunctionArgs) {
+
+        if(args.callback){
```
mikhailshilkov, line 83:
```suggestion
        if (args.callback) {
```

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+            throw new Error("Durable entity functions need to use the [callbackFactory]");
+        }
+        
+        const trigger = {
+            name: "context",
+            type: "entityTrigger",
+            direction: "in"
```
mikhailshilkov, line 90:
```suggestion
            direction: "in",
```

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class DurableEntityClientInputBindingSettings implements mod.InputBindingSettings {
+    binding: pulumi.Input<mod.InputBindingDefinition>;
+    settings: pulumi.Input<{ [key: string]: any; }>;
+
+    constructor(name: string) {
+        this.binding = {
+            name,
+            type: "durableClient",
+            direction: "in"
```
mikhailshilkov, line 119:
```suggestion
            direction: "in",
```

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+
+export class DurableOrchestrationClientInputBindingSettings implements mod.InputBindingSettings {
+    binding: pulumi.Input<mod.InputBindingDefinition>;
+    settings: pulumi.Input<{ [key: string]: any; }>;
+
+    constructor(name: string) {
+        this.binding = {
+            name,
+            type: "orchestrationClient",
+            direction: "in"
```
mikhailshilkov, line 105:
```suggestion
            direction: "in",
```

examples/durable-functions/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                const client = df.getClient(con);
+                client.startNew("orch", req.params.id);
+
+                return client.waitForCompletionOrCreateCheckStatusResponse(req, req.params.id, 5000, 100);
+            },
+            inputs: [new azure.appservice.DurableOrchestrationClientInputBindingSettings("starter")]
+        }),
+        new azure.appservice.HttpFunction("viewCount", {
+            route: "views",
+            callback: async (con, req) => {
```
mikhailshilkov, line 54: Let's call it `context` everywhere

examples/durable-functions/index.ts
```diff
@@ -0,0 +1,69 @@
...
+                    status: 200,
+                    body: stateResponse.entityState
+                };
+            },
+            inputs: [new azure.appservice.DurableEntityClientInputBindingSettings("client")]
+        })
+    ]
+});
+
+export const url = app.endpoint;
```
mikhailshilkov, line 69: Do you mind adding a short instruction on how to test the deployed functions?

examples/durable-functions/index.ts
```diff
@@ -0,0 +1,69 @@
+// Copyright 2016-2017, Pulumi Corporation.  All rights reserved.
+
+import * as azure from "@pulumi/azure";
+import * as df from 'durable-functions';
+
+const resourceGroup = new azure.core.ResourceGroup("chedDurable", {
```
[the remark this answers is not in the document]
ChristianEder: oh my. I'll rename

examples/durable-functions/index.ts
```diff
@@ -0,0 +1,69 @@
...
+    location: azure.Locations.WestEurope
+});
+
+var app = new azure.appservice.MultiCallbackFunctionApp("durable", {
+    resourceGroup,
+    functions: [
+        new azure.appservice.DurableActivityFunction("SayHello", {
+            activityInputName: "name",
+            callback: (context) => {
+                context.done(null, "Hello, " + context.bindings.name);
```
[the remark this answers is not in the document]
ChristianEder: should be, will change

examples/examples_test.go
```diff
@@ -26,6 +26,16 @@ func TestAccMinimal(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccDurableFunctions(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: path.Join(getCwd(t), "durable-functions"),
+			//RunUpdateTest: true, // temporarily disabled as upgrade tests are not able to run until we release 2.0.0
```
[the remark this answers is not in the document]
ChristianEder: I will change this throughout the file

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import * as pulumi from "@pulumi/pulumi";
+import * as mod from ".";
+
+export interface DurableOrchestratorFunctionContext extends mod.Context<mod.Result> {
+    df: any
```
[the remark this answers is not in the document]
ChristianEder: Probably not really - unless I not either add a dependency to the 'durable-functions' package (I would assume thats not a solution), or try to replicate the df interface. The package defines the context as

`export interface IOrchestrationFunctionContext extends Context {
    df: DurableOrchestrationContext;
}`

sdk/nodejs/appservice/zMixins_durable.ts
```diff
@@ -0,0 +1,123 @@
...
+            direction: "in"
+        } as mod.InputBindingDefinition;
+
+        super(name, trigger, args);
+    }
+}
+
+export interface DurableActivityFunctionContext<TActivityInputBinding> extends mod.Context<mod.Result> {
+    bindings: {
+        [key: string]: any;
```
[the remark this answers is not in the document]
ChristianEder: hmm I probably had a reason when I added that but it compiles without it soo... I'll remove it.

examples/durable-functions/index.ts
ChristianEder: added a comment explaining how to call the durable clients via http

ChristianEder:
I worked through all your comments, resolved the "easy" ones - do you want to resolve the others / give more feedback?

examples/durable-functions/index.ts
```diff
@@ -1,19 +1,17 @@
...
+const resourceGroup = new azure.core.ResourceGroup("durable-example");
 
 var app = new azure.appservice.MultiCallbackFunctionApp("durable", {
     resourceGroup,
     functions: [
         new azure.appservice.DurableActivityFunction("SayHello", {
             activityInputName: "name",
-            callback: (context) => {
-                context.done(null, "Hello, " + context.bindings.name);
+            callback: async(context) => {
```
mikhailshilkov, line 13:
```suggestion
            callback: async context => {
```

examples/examples_test.go
```diff
@@ -26,11 +26,21 @@ func TestAccMinimal(t *testing.T) {
 	integration.ProgramTest(t, &test)
 }
 
+func TestAccDurableFunctions(t *testing.T) {
+	test := getJSBaseOptions(t).
+		With(integration.ProgramTestOptions{
+			Dir: path.Join(getCwd(t), "durable-functions"),
+			RunUpdateTest: true
```
mikhailshilkov, line 33: You have to put `,` at the end of each line (see build errors). Sorry, I didn't realize that 2.0 thing was all over the place, feel free to revert that changes - they shouldn't be your problem.

ChristianEder:
> > unless I not either add a dependency to the 'durable-functions' package (I would assume thats not a solution),
> 
> 
> 
> We could add those. We already have dependencies e.g. on `azure-functions-ts-essentials` and `azure-eventgrid`.

From the perspective of a durable function developer, adding the correct type will not add any benefit - because the method containing the durable function itself is wrapped inside a `df.orchestrator`call and gets the correct typing anyway. This type here is only required so that the callback returned by the `df.orchestrator` call gets accepted as a valid callback by the compiler. 

If you‘d prefer having the right types defined anyway, I can add them - but wouldn’t that mean the package being deployed to _any_ function app, even those that don’t use durable functions?
```

mikhailshilkov:
Alright, let's keep `any` then. Thanks for the explanation!

ChristianEder:
So - build is back to green, all comments resolved. *phew* :-)

PR #475, opened 2020-03-07
Revert "Upgrade to v2.0.0 of the AzureRM Terraform Provider"

leezen:
Reverts pulumi/pulumi-azure#471


leezen:
Looks like the test itself needs to be changed. Closing.

PR #476, opened 2020-03-07
Remove deprecated property from IoT Hub

mikhailshilkov:


mikhailshilkov:
The failure is unrelated, I'll investigate separately

PR #479, opened 2020-03-07
Retire ZipBlob in favor of a better Blob

mikhailshilkov:
Use the new major version 2.0 to retire an old workaround of `ZipBlob` resource. Make `Blob` support assets and archives instead.

The changes:
- `azure.storage.ZipBlob` is removed
- `azure.storage.Blob` gets a new property `content` instead of the property `source`

Examples of the new `Blob` usage:

Upload an HTML file. Pulumi will watch the content changes and re-upload if needed (TypeScript):

```ts
const sampleFile = new azure.storage.Blob("test.html", {
    storageAccountName: storageAccount.name,
    storageContainerName: storageContainer.name,
    type: "Block",
    content: new pulumi.asset.FileAsset("./test.html"),
    contentType: "text/html",
});
```

Zip the contents of a folder and upload the archive as a blob (C#):

```cs
var blob = new Blob("zip", new BlobArgs
{
    StorageAccountName = storageAccount.Name,
    StorageContainerName = container.Name,
    Type = "Block",
    Content = new FileArchive("wwwroot")
});
```


mikhailshilkov:
@lukehoban Marked this issue and the note in the changelog as breaking. Expanded the PR description with more details and two examples.

resources.go
```diff
@@ -1073,6 +1072,12 @@ func Provider() tfbridge.ProviderInfo {
 							return strings.ToLower(name)
 						},
 					}),
+					"source": {
+						Name: "content",
+						Asset: &tfbridge.AssetTranslation{
+							Kind: tfbridge.FileAsset,
+						},
```
lukehoban: @mikhailshilkov 

One thing I realized after reviewing some code that currently uses `ZipBlob` is that the pair of changes here leads to there being no code that can be written which works on both 1.0 and 2.0.  That's a particualrly unfortunate style of breaking change.  

I wonder if we might be able to have a cleaner transition is we instead:
1. Kept this property named `source`
2. Had it support an `AltType` of `string`.  

I'm not sure we are able to support (2) yet today unfortunately.  

Other thoughts?  Or is the current break truly the right thing to do?

mikhailshilkov: You are right.

We could keep the property named `source` instead of `content`. This is not aligned to what we did in AWS, but doesn't seem important. Interestingly, it seems that we can assign a `string` to this property, at least in TypeScript:

``` ts
const sampleFile = new azure.storage.Blob("test.html", {
    name: "test.html",
    storageAccountName: storageAccount.name,
    storageContainerName: storageContainer.name,
    type: "Block",
    content: "./test.html",
    contentType: "text/html",
});
```

However, we probably don't want people to do that, because in this case Pulumi will not watch the content changes in the file as opposed to `FileAsset`.

Now, in `1.0` one couldn't upload a folder archive with `Blob`: you had to use `ZipBlob`. If we want, we could keep `ZipBlob` around in 2.0 (marked as deprecated) which would help with the transition of folder archives.

So, the most compatible way seems to be:
1. Rename `content` back to `source`. Basically, give up we will ever change it.
2. Restore `ZipBlob` as a deprecated resource (well, we still don't have deprecation 1st class support, do we?).

lukehoban: I am okay with doing both (1) and (2) in this list. It's less progress on migrating forward, but helps ensure this is not too disruptive for now.

Issue #492, opened 2020-03-13
Webhook classes are misspelled as "Webook"


kylehayes:
All related webhook classes seem to have a misspelling as "webook" for example: https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/containerservice/#RegistryWebook

stack72:
Hi @kylehayes 

I’m so sorry about this - I’ll have this fixed ASAP and try and make in next release

@stack72

PR #493, opened 2020-03-20
Rename containerService.Webook to containerService.Webhook

stack72:
Fixes: #492

Also adds some missing data sources


mikhailshilkov:
The old resource doesn't seem to be marked as deprecated. Do we still not support that?

stack72:
@mikhailshilkov we do not - can you open an issue in the bridge repo to add that?

mikhailshilkov:
@stack72 Opened https://github.com/pulumi/pulumi-terraform-bridge/issues/133

stack72:
thanks!!

PR #494, opened 2020-03-23
Upgrade to v2.2.0 of the AzureRM Terraform Provider

stack72:


resources.go
```diff
@@ -1228,11 +1236,14 @@ func Provider() tfbridge.ProviderInfo {
 			// IoT Central
 			"azurerm_iotcentral_application": {Tok: azureResource(azureIotCentral, "Application")},
 
-			// HPC
+			//
```
mikhailshilkov: Is this intentional?

resources.go
```diff
@@ -797,6 +799,11 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_cosmosdb_gremlin_database":   {Tok: azureResource(azureCosmosDB, "GremlinDatabase")},
 			"azurerm_cosmosdb_gremlin_graph":      {Tok: azureResource(azureCosmosDB, "GremlinGraph")},
 
+			// Cost Management
+			"azurerm_cost_management_export_resource_group": {
+				Tok: azureResource(azureCostManagement, "ExportResourceGroup"),
```
mikhailshilkov: `ExportResourceGroup` sounds a bit weird... Maybe `ResourceGroupExport`?

resources.go
stack72: nope - fixed

resources.go
stack72: fixed :)

PR #496, opened 2020-03-23
Update to use new module system 

EvanBoyle:
1. Consume new /sdk and /pkg modules from pulumi/pulumi
2. Split this repo into a multi-module repo so that `pulumi/pulumi-azure/sdk` can be imported without bringing in core pulumi/pulumi/pkg dependencies. 

This still has a local replace on tfbridge as this change is blocked on https://github.com/pulumi/pulumi-terraform-bridge/pull/134


stack72:
@EvanBoyle this looks great! thanks for this - works for me locally as well

PR #505, opened 2020-03-30
Bump modules to v2 structure

stack72:
This is required so that users can run go get and the version without
needing to know the individual Git SHA

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -53,10 +53,10 @@ generate_schema:: tfgen
 
 provider::
 	go generate ${PROJECT}/provider/cmd/${PROVIDER}
-	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/pkg/version.Version=${VERSION}" ${PROJECT}/provider/cmd/${PROVIDER}
+	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/v2/pkg/version.Version=${VERSION}" ${PROJECT}/provider/v2/cmd/${PROVIDER}
 
 tfgen::
-	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/pkg/version.Version=${VERSION}" ${PROJECT}/provider/cmd/${TFGEN}
+	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/v2/pkg/version.Version=${VERSION}" ${PROJECT}/provider/v2/cmd/${TFGEN}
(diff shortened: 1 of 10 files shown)


Makefile
```diff
@@ -53,10 +53,10 @@ generate_schema:: tfgen
 
 provider::
 	go generate ${PROJECT}/provider/cmd/${PROVIDER}
-	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/pkg/version.Version=${VERSION}" ${PROJECT}/provider/cmd/${PROVIDER}
+	cd provider && go install -ldflags "-X github.com/pulumi/pulumi-azure/provider/v2/pkg/version.Version=${VERSION}" ${PROJECT}/provider/cmd/${PROVIDER}
```
jen20: Probably worth checking the version is propagated correctly here.

stack72: It is not - it *should* be but it's not :/

stack72: ok, fixed :)

EvanBoyle:
Ah, I remember now. The examples don't need a local replace (or require) because we don't test any of the go examples in this repo (or GCP). Nevermind my previous comment.

scripts/publish-plugin.sh
```diff
@@ -17,7 +17,7 @@ if [ "$(go env GOOS)" = "windows" ]; then
 fi
 
 (cd "${ROOT}/provider" && go build \
-   -ldflags "-X github.com/pulumi/pulumi-azure/provider/pkg/version.Version=${VERSION}" \
+   -ldflags "-X github.com/pulumi/pulumi-azure/provider/pkg/v2/version.Version=${VERSION}" \
    -o "${WORK_PATH}/pulumi-resource-azure${BIN_SUFFIX}" \
    "${ROOT}/cmd/pulumi-resource-azure")
```
EvanBoyle: Does this same thing apply here with the package being updated to "${ROOT}/cmd/pulumi-resource-azure?

stack72: yes it will - you are correct

stack72: actually no, that *is* a path as it;s a go build

```
▶ ./scripts/publish-plugin.sh
can't load package: package ./v2/cmd/pulumi-resource-azure: cannot find package "." in:
	/Users/stack72/code/go/src/github.com/pulumi/pulumi-azure/provider/v2/cmd/pulumi-resource-azure
```

PR #513, opened 2020-04-06
Re-Enable linting of provider as part of CI

stack72:
Related: pulumi/pulumi#3834


mikhailshilkov:
> fatal error: runtime: out of memory

Is that because of linting or unrelated?

stack72:
Handling as part of move to GHA

PR #520, opened 2020-04-14
Merging feature-2.0 back to master

stack72:


stack72:
2 tests fail in this branch but both have been disabled in master - so merging!

PR #521, opened 2020-04-14
Python example to create an AKS cluster

AdminTurnedDevOps:
This example will create an AKS cluster. I left out any `sys` runtimes as I wasn't sure if the Pulumi team would prefer any parameter runtimes to be passed in or not.


mikhailshilkov:
Hi @AdminTurnedDevOps, thanks for opening this PR.

We usually put our public examples to the central repo here: https://github.com/pulumi/examples
The examples in provider repos are meant mostly for integration testing.

There's already an AKS/Python example in https://github.com/pulumi/examples/tree/master/azure-py-aks. Do you feel like that example is not enough to get started with AKS and python? If so, consider contributing to improve it.

Thank you!

AdminTurnedDevOps:
That makes sense! Thanks. I was looking under the Azure examples, but I couldn’t find it. Now I see it was in the main examples page.

Perhaps one example under the Azure portion?

PR #554, opened 2020-05-07
Don't run update test for non-released code

mikhailshilkov:
I keep forgetting the modes of our tests, but I think update tests aren't allowed until a piece of functionality is released. Switch it off for now to fix https://travis-ci.com/github/pulumi/pulumi-azure/builds/164557586#L3596


mikhailshilkov:
Self-merged to fix the master build

PR #555, opened 2020-05-08
Fix `error404document` mapping on azure.storage.Account

### LINKED ISSUE #512: Azure Storage Static Website: invalid or unknown key: error404_document
I'm using a C# project to create an Azure Storage static website but adding a value for `Error404Document` results in the error:
```
Diagnostics:
  azure:storage:Account (x):
    error: azure:storage/account:Account resource 'x' has a problem: static_website.0: invalid or unknown key: error404_document
```
Removing just the `Error404Document` property results in a successful deployment.

Here's an example resource:
```
            var storageAccount = new Account("x", new AccountArgs
            {
                ResourceGroupName = resourceGroup.Name,
                EnableHttpsTrafficOnly = true,
                AccountReplicationType = "LRS",
                AccountTier = "Standard",
                AccountKind = "StorageV2",
                AccessTier = "Cool",
                StaticWebsite = new AccountStaticWebsiteArgs
                {
                    IndexDocument = "index.html",
                    Error404Document = "index.html",
                },
            });
```
And I'm using version 2.4.1 of the Azure package:
```
    <PackageReference Include="Pulumi.Azure" Version="2.4.1-preview" />
```

lukehoban:
The bridge's mapping produces the wrong name for the `error_404_document` property.   We can override this to ensure we map to the correct name in the upstream schema.

Fixes #512.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Upgrade to pulumi-terraform-bridge v2.2.0
+* Fix `error404document` mapping on `azure.storage.Account`
 
 ---
 
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -1099,7 +1099,23 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_signalr_service": {Tok: azureResource(azureSignalr, "Service")},
 
 			// Storage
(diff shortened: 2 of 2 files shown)


mikhailshilkov:
Build error is unrelated and has been fixed in master

stack72:
The test that is broken is broken in master

PR #562, opened 2020-05-10
Add function to compute CosmosDB Connection string

davidobrien1985:
Provider is currently missing a way to get the primary connection string from a SQL CosmosDB account.
This is my first try at the NodeJS implementation.
My laptop didn't want to run `make` so I hope this is an acceptable first attempt :)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ CHANGELOG
 * Fix `error404document` mapping on `azure.storage.Account`
 * Upgrade to v2.9.0 of the AzureRM Terraform Provider
 * Update ArchiveFunctionApp to pass correct params to FunctionApp
+* Add NodeJS function to build CosmosDB Primary Connection String for SQL Accounts
 
 ---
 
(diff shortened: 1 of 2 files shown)


mikhailshilkov:
Hey @davidobrien1985 
Isn't [conectionStrings](https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/cosmosdb/account.ts#L49) doing this already? I believe we use the first one in Azure Funcitons [here](https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/cosmosdb/zMixins.ts#L217-L218).

davidobrien1985:
For whatever reason it only does that for Accounts that are not SQL.
https://www.pulumi.com/docs/reference/pkg/nodejs/pulumi/azure/cosmosdb/#Account-connectionStrings

Sent from Outlook Mobile<https://aka.ms/blhgte>

________________________________

mikhailshilkov:
Sorry I left this PR out. I think our current stance is not to introduce custom functionality in this provider and reuse whatever comes from TF. We'll focus on making Azure NextGen great.

PR #574, opened 2020-05-15
Upgrade to v2.11.0 of the AzureRM Terraform Provider

stack72:
Fixes: #573 
Fixes: #579


sdk/dotnet/MariaDB/GetMariaDbServer.cs
```diff
@@ -50,10 +50,6 @@ public sealed class GetMariaDbServerResult
         /// </summary>
         public readonly string AdministratorLogin;
         /// <summary>
-        /// The password associated with the `administrator_login` for the MariaDB Server.
-        /// </summary>
-        public readonly string AdministratorLoginPassword;
```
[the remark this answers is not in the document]
stack72: This is 100% a breaking change that happened upstream. FWIW, this was *never* set in the schema - passwords were never returned from the API

provider/cmd/pulumi-resource-azure/schema.json
```diff
@@ -100547,6 +101272,8 @@
                 },
                 "type": "object",
                 "required": [
+                    "identities",
+                    "location",
```
lukehoban, line 101276: No action needed here - but it is interesting to observe that new required fields would in a traditional sense also be a breaking change - but seems to not be considered one here by the upstream provider?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ CHANGELOG
 
 ## HEAD (Unreleased)
 * Upgrade to pulumi-terraform-bridge v2.3.3
+* Update to v2.10.0 of the AzureRM Terraform Provider
```
lukehoban, line 6: BTW - as a general thing - and I expect this may add some work required - it would be very nice to hyperlink to changelog entry of the upstream provider here.

provider/resource.go
```diff
@@ -1504,6 +1515,7 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_eventhub_namespace_authorization_rule": {
 				Tok: azureDataSource(azureEventHub, "getNamespaceAuthorizationRule"),
 			},
+			"azurerm_eventhub":                        {Tok: azureDataSource(azureEventHub, "getEventhub")},
```
mikhailshilkov, line 1518: Should be `getEventHub`

PR #575, opened 2020-05-18
Use FunctionAppIdentity for packagedfunctionapp args

adnang:
This is to give all the "specialised" function app resources access to UserAssigned identity, as a normal function app would : https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/appservice/functionApp.ts#L427. This probably generalizes to all the other function app args too.


adnang:
@mikhailshilkov thanks, after pulling the pulumi:master down on top of my branch and merging the changelog, the build seems to have failed https://travis-ci.com/github/pulumi/pulumi-azure/builds/168064425. Is this unrelated?

stack72:
Hey @adnang 

yes, that was unrelated - this was just due to a failure from the GitHub API

Thanks for rebasing this for us

Paul

PR #577, opened 2020-05-20
Mark client_secret and client_certificate_password as Secret 

### LINKED ISSUE #576: Sensitive `ARM_` env vars get serialized into state file (and in plain text)
Sensitive ARM env vars like `ARM_CLIENT_SECRET` will currently get picked up from the environment and stored into the statefile in plain text.  Two issues:
1. It is not clear these (and even some of the non-sensitive env var) belong in the state in the first place
2. If they do need to be in the state, they should be marked as secret and thus encrypted by default

lukehoban:
Ensure that the default values for `client_secret` and `client_certificate_password` picjed up from the environment will be marked as secrets (and thus encrypted in the state file).

It remains unclear long-term that it is correct to be embedding these values in the state file at all - in many cases they will be short lived credentials or even entirely unused for a particular deployment (superseded by other configuration for a provider).  However, for now, we should at least ensure they are marked as secrets so that sensitive data doesn't get pulled into the state file in plain text.  We will revisit what gets stored as provider inputs as part of https://github.com/pulumi/pulumi-terraform-bridge/issues/14.

Fixes #576.


lukehoban:
After discussion with @pgavlin offline, it seems we may want to instead provide a way to mark these inputs as secret as an immediate solution.  We don't necessarily want to lose all of this data from provider inputs, as it will be important once we address https://github.com/pulumi/pulumi-terraform-bridge/issues/14 to have it.  We may need to further rethink this - but can address the immediate concern by focusing on ensuring sensitive data is marked as secret and encrypted in the state file.

lukehoban:
After some offline discussion - we are planning to go back to the original approach here - now tracked in https://github.com/pulumi/pulumi-azure/pull/583.

PR #583, opened 2020-05-26
Avoid baking config from environment into checkpoint

### LINKED ISSUE #576: Sensitive `ARM_` env vars get serialized into state file (and in plain text)

lukehoban:
Like in https://github.com/pulumi/pulumi-aws/pull/894, we will avoid baking any provider configuration from environment variables into the checkpoint file for any potentially transient configuration (like credentials) that is not part of the "identity" of the provider.

This notion is still a little loosely defined, and will be more precisely defined for each provider as part of https://github.com/pulumi/pulumi-terraform-bridge/issues/14.  For now though, there is no reason to include these, and indeed they can lead to (a) leaking of sensitive data from the environment into the state file and (b) storing invalid (expired) data in the checkpoint which unintentionally gets used for future destroy/refresh operations.

Fixes #576.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,10 @@ CHANGELOG
   **PLEASE NOTE:** `azure.mariaDb.getMariaDbServer` 'administratorLoginPassword' has been removed. This was
   a property that was never available from the Azure API so was never accessible.
 * Use FunctionAppIdentity for packagedfunctionapp args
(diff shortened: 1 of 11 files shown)


lukehoban:
Tracking doing the same for all other providers at https://github.com/pulumi/pulumi-terraform-bridge/issues/209.

PR #587, opened 2020-05-29
Upgrade to v2.12.0 of the AzureRM Terraform Provider

stack72:
Fixes: #585


provider/resource.go
```diff
@@ -606,12 +607,13 @@ func Provider() tfbridge.ProviderInfo {
...
-			"azurerm_dev_test_schedule":                {Tok: azureResource(azureDevTest, "Schedule")},
-			"azurerm_dev_test_virtual_network":         {Tok: azureResource(azureDevTest, "VirtualNetwork")},
-			"azurerm_dev_test_windows_virtual_machine": {Tok: azureResource(azureDevTest, "WindowsVirtualMachine")},
+			"azurerm_dev_test_lab":                         {Tok: azureResource(azureDevTest, "Lab")},
+			"azurerm_dev_test_linux_virtual_machine":       {Tok: azureResource(azureDevTest, "LinuxVirtualMachine")},
+			"azurerm_dev_test_policy":                      {Tok: azureResource(azureDevTest, "Policy")},
+			"azurerm_dev_test_schedule":                    {Tok: azureResource(azureDevTest, "Schedule")},
+			"azurerm_dev_test_virtual_network":             {Tok: azureResource(azureDevTest, "VirtualNetwork")},
+			"azurerm_dev_test_windows_virtual_machine":     {Tok: azureResource(azureDevTest, "WindowsVirtualMachine")},
+			"azurerm_dev_test_global_vm_shutdown_schedule": {Tok: azureResource(azureDevTest, "GlobalVmShutdownSchedule")},
```
mikhailshilkov, line 616:
```suggestion
			"azurerm_dev_test_global_vm_shutdown_schedule": {Tok: azureResource(azureDevTest, "GlobalVMShutdownSchedule")},
```

provider/resource.go
```diff
@@ -1018,8 +1020,11 @@ func Provider() tfbridge.ProviderInfo {
 					Source: "express_route_circuit_authorization.html.markdown",
 				},
 			},
-			"azurerm_express_route_gateway":          {Tok: azureResource(azureNetwork, "ExpressRouteGateway")},
-			"azurerm_nat_gateway":                    {Tok: azureResource(azureNetwork, "NatGateway")},
+			"azurerm_express_route_gateway": {Tok: azureResource(azureNetwork, "ExpressRouteGateway")},
+			"azurerm_nat_gateway":           {Tok: azureResource(azureNetwork, "NatGateway")},
+			"azurerm_nat_gateway_public_ip_association": {
+				Tok: azureResource(azureNetwork, "NatGatewayPublicIpAssociation"),
```
mikhailshilkov, line 1026: Also `Ip` should in theory be `IP` but we are already inconsistent with it, so I don't have a good answer what to do here.

provider/resource.go
stack72: fixed!

provider/resource.go
stack72: i would suggest we open another issue to track the inconsistencies and then we can fix those up at a later time

PR #598, opened 2020-06-11
Upgrade to v2.14.0 of the AzureRM Terraform Provider

stack72:
Fixes: #597


sdk/dotnet/ContainerService/GetKubernetesCluster.cs
```diff
@@ -52,15 +52,6 @@ public sealed class GetKubernetesClusterArgs : Pulumi.InvokeArgs
...
         public string Name { get; set; } = null!;
 
-        /// <summary>
-        /// If the cluster has the Kubernetes API only exposed on internal IP addresses.                           
-        /// </summary>
-        [Input("privateClusterEnabled")]
-        public bool? PrivateClusterEnabled { get; set; }
-
-        [Input("privateLinkEnabled")]
-        public bool? PrivateLinkEnabled { get; set; }
```
mikhailshilkov, line 62: That looks like a breaking change, what happened to these two?

stack72: According to the CHANGELOG in AzureRM Provider:

```
Data Source: azurerm_kubernetes_cluster
bug fix: fixing an issue where some read-only fields were unintentionally marked as user-configurable.
```

These are those specific fields!

https://github.com/terraform-providers/terraform-provider-azurerm/pull/7233/files#diff-ceb52ca7b573864a4c727e76b6e4c09cL234

PR #613, opened 2020-06-23
pinning virtualenv version

jaxxstorm:


stack72:
I've taken care of this in master already

PR #623, opened 2020-06-30
Do not install the Azure provider.

pgavlin:
This is no longer necessary for example conversion.


Makefile
```diff
@@ -62,7 +62,6 @@ tfgen::
 
 install_plugins::
 	[ -x "$(shell which pulumi)" ] || curl -fsSL https://get.pulumi.com | sh
-	pulumi plugin install resource $(PACK) 3.9.0
```
stack72, line 65: This is a big success!!! I will remove this from all the providers as I am revving the others :)

PR #638, opened 2020-07-17
Upgrade to v2.19.0 of the AzureRM Terraform Provider

stack72:
Fixes: #637


sdk/dotnet/Monitoring/MetricAlert.cs
```diff
@@ -144,7 +156,7 @@ public partial class MetricAlert : Pulumi.CustomResource
         /// A set of strings of resource IDs at which the metric criteria should be applied.
         /// </summary>
         [Output("scopes")]
-        public Output<string> Scopes { get; private set; } = null!;
+        public Output<ImmutableArray<string>> Scopes { get; private set; } = null!;
```
mikhailshilkov, line 159: That's a breaking change. Is that a bug fix?

stack72: Fraid not :/ https://github.com/terraform-providers/terraform-provider-azurerm/pull/7159

mikhailshilkov: @lukehoban does that pass our upgrade policy gate?

lukehoban: Unfortunately we still don't have a great answer around these.  

This is a bug fix in the sense that the platform does allow multiple items here, and previously Terraform/Pulumi only allowed one.  It is not a breaking change for Terraform users.

We could:
1. Override `MaxItems: 1` to keep the restriction in Pulumi to avoid a breaking change.
2. Accept the breaking change and document it.

Given that this is not a heavily used resource currently, let's do (2).

Going forward, we should discuss whether there is anything more structurally better we can do to address cases where upstream removes `MaxItems: 1` annotations - which are not breaking changes upstream but are for Pulumi.

Issue #644, opened 2020-07-28
Naming restrictions for API Management APIs are too strict


Gerrit-K:
Similarly to #159 the naming restrictions for the `API` resource within the API Management are too strict. I get an error if I exceed 16 characters (24 - 8 random), which I believe is the (conservative) default value for all Pulumi resources. 

According to [the Azure docs](https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftapimanagement) however, the maximum for this type should be 256. I can confirm that it's possible to manually create APIs with names of more than 24 characters from within the portal.

stack72:
Hi @Gerrit-K 

Ok, now I understand what you mean, I am going to try and file a PR upstream to the Terraform provider we derive this provider from

I will also undo the change originally associated with this issue

Paul

stack72:
ok, on further investigation @Gerrit-K 

The offending line isn't upstream it's actually https://github.com/pulumi/pulumi-azure/blob/master/provider/resource.go#L1947 🤦 

Looking at the impact of changing this now

Paul

Gerrit-K:
@stack72 
Hey Paul,

well, yes, that's something from Pulumi and I believe I've read somewhere that this limit - while being very conservative - is necessary to avoid errors for cases when the limits are unknown or subject to change. The linked docs however describe many limits in great detail and I've seen a couple of resources in Pulumi where these limits are not accounted for. Ideally I would love to see all of them implemented, but I can also imagine that this might be a better fit for the upstream terraform provider (does it have those limits as well?).

stack72:
Hi @Gerrit-K 

This PR #648 will fix the issue - you can see, it's quite easy to change these restrictions 

Hope this helps

Paul

PR #645, opened 2020-07-28
Relax naming conventions for APIManagement Service name to allow 256 character names

stack72:
Fixes: #644

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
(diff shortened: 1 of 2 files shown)


Gerrit-K:
@stack72 Thanks for the quick implementation, but I'm afraid there was a misunderstanding. In #644 I was not referring to API Management `Service`s (for which the existing limit of 50 characters actually was fine, according to the docs), but instead to the `API` resource in the same namespace (which belongs to the scope of a `Service`, one level down in hierarchy).

mikhailshilkov:
Yeah, this is wrong, @stack72 please revert

PR #648, opened 2020-07-30
Relax naming conventions for APIManagement API name to allow 256 character names

stack72:
Fixes: #644

### DIFF
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -299,7 +299,18 @@ func Provider() tfbridge.ProviderInfo {
 					}),
 				},
 			},
-			"azurerm_api_management_api":                         {Tok: azureResource(azureAPIManagement, "Api")},
+			"azurerm_api_management_api": {
+				Tok: azureResource(azureAPIManagement, "Api"),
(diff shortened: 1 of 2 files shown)


provider/resource.go
```diff
@@ -299,7 +299,18 @@ func Provider() tfbridge.ProviderInfo {
...
 				},
 			},
-			"azurerm_api_management_api":                         {Tok: azureResource(azureAPIManagement, "Api")},
+			"azurerm_api_management_api": {
+				Tok: azureResource(azureAPIManagement, "Api"),
+				Fields: map[string]*tfbridge.SchemaInfo{
+					// Max length of an API Management API name is 256.
+					// Source: https://docs.microsoft.com/en-us/azure/azure-resource-manager/management/resource-name-rules#microsoftapimanagement
+					azureName: tfbridge.AutoNameWithCustomOptions(azureName, tfbridge.AutoNameOptions{
+						Separator: "-",
```
mikhailshilkov, line 308: Why is the separator `-` here and empty in the `Service` resource? Isn't this going to break existing autonamed resources?

PR #669, opened 2020-08-19
Upgrade to Pulumi v2.9.0

stack72:


CHANGELOG.md
```diff
@@ -2,7 +2,8 @@ CHANGELOG
 =========
 
 ## HEAD (Unreleased)
-_(none)_
+* Upgrade to Pulumi v2.9.0
```
justinvp, line 5:
```suggestion
* Upgrade to Pulumi v2.9.0, which adds type annotations and input/output classes to Python
```

sdk/python/pulumi_azure/network/get_public_ips.py
```diff
@@ -60,7 +108,7 @@ def get_public_i_ps(allocation_type=None,attached=None,name_prefix=None,resource
     import pulumi
     import pulumi_azure as azure
 
-    example = azure.network.get_public_i_ps(attached=False,
+    example = azure.network.get_public_ips(attached=False,
```
justinvp, line 111: This would also be a breaking change. Working on a codegen fix that special-cases this to use the old name.

justinvp: https://github.com/pulumi/pulumi/pull/5202

justinvp: We'll need to update the `go.mod` require for pulumi/pulumi in this PR to reference this commit: https://github.com/pulumi/pulumi/commit/62925433421370e32e614f2d2647f577673148bc

PR #674, opened 2020-08-24
Add webserver-py and webserver-py-old examples

justinvp:
webserver-py is a Python example based on the Node.js example that uses the new input/output classes. webserver-py-old uses dicts instead of the input/output classes, to ensure the old way of setting/getting nested data continues to work as expected.

Also, some minor cleanup while making changes to these test files: switched to using `filepath.Join`, rather than `path.Join`.


stack72 (approved):
<3

PR #730, opened 2020-10-09
Upgrade to v2.31.1 of the AzureRM Terraform Provider

stack72:
Fixes: #726
Fixes: #727


provider/resource.go
```diff
@@ -43,88 +43,89 @@ const (
...
+	azureCore                  = "Core"                  // Base Resources
+	azureCosmosDB              = "CosmosDB"              // Cosmos DB
+	azureCostManagement        = "CostManagement"        // CostManagement
+	azureDashboard             = "Dashboard"             // Dashboard
+	azureDatabaseMigration     = "DatabaseMigration"     // Database Migration
+	azureDataFactory           = "DataFactory"           // Data Factory
+	azureDatalake              = "DataLake"              // Data Lake
+	azureDataShare             = "DataShare"             // DataShare
+	azureDataBricks            = "DataBricks"            // DataBricks
+	azureDesktopVirtualization = "DesktopVirtualization" // DesktopVirtualization
```
mikhailshilkov, line 73: I believe the service is "Virtual Desktop" - the same as the TF name. Why did you pick "Desktop Virtualization"?

stack72: I picked it as that was in the TF module - https://github.com/terraform-providers/terraform-provider-azurerm/pull/8605/files#diff-d626fff06c84ff73c96b49db331f82ceR33

Will change it now though :)

mikhailshilkov: I think they are right, the same name is used in Open API specs: https://github.com/Azure/azure-rest-api-specs/tree/master/specification/desktopvirtualization/resource-manager

Let's go ahead with the current version, sorry for the confusion!

stack72: Ah I forgot to change this back - I will do this now

mikhailshilkov (approved):

PR #754, opened 2020-11-05
Upgrade to v2.35.0 of the AzureRM Terraform Provider

stack72:
Fixes: #753


sdk/dotnet/Network/GetGatewayConnection.cs
```diff
@@ -128,7 +128,7 @@ public sealed class GetGatewayConnectionResult
         /// A mapping of tags to assign to the resource.
         /// </summary>
         public readonly ImmutableDictionary<string, string> Tags;
-        public readonly Outputs.GetGatewayConnectionTrafficSelectorPolicyResult TrafficSelectorPolicy;
+        public readonly ImmutableArray<Outputs.GetGatewayConnectionTrafficSelectorPolicyResult> TrafficSelectorPolicies;
```
mikhailshilkov, line 131: Looks like a breaking change. Can we MaxItems=1 it?

sdk/dotnet/Policy/Inputs/PolicySetDefinitionPolicyDefinitionReferenceArgs.cs
```diff
@@ -19,15 +19,15 @@ public sealed class PolicySetDefinitionPolicyDefinitionReferenceArgs : Pulumi.Re
...
         [Input("parameters")]
-        private InputMap<object>? _parameters;
+        private InputMap<string>? _parameters;
 
         /// <summary>
         /// Parameters for the policy set definition. This field is a json object that allows you to parameterize your policy definition.
         /// </summary>
         [Obsolete(@"Deprecated in favour of `parameter_values`")]
-        public InputMap<object> Parameters
+        public InputMap<string> Parameters
```
mikhailshilkov, line 28: This is technically breaking, although I suspect it became "correct" now.

sdk/dotnet/Network/GetGatewayConnection.cs
stack72: Fixed!

sdk/go/azure/network/virtualHub.go
```diff
@@ -64,24 +64,20 @@ type VirtualHub struct {
 	ResourceGroupName pulumi.StringOutput `pulumi:"resourceGroupName"`
 	// One or more `route` blocks as defined below.
 	Routes VirtualHubRouteArrayOutput `pulumi:"routes"`
+	// The sku of the Virtual Hub. Possible values are `Basic` and `Standard`. Changing this forces a new resource to be created.
+	Sku pulumi.StringPtrOutput `pulumi:"sku"`
 	// A mapping of tags to assign to the Virtual Hub.
 	Tags pulumi.StringMapOutput `pulumi:"tags"`
 	// The ID of a Virtual WAN within which the Virtual Hub should be created. Changing this forces a new resource to be created.
-	VirtualWanId pulumi.StringOutput `pulumi:"virtualWanId"`
+	VirtualWanId pulumi.StringPtrOutput `pulumi:"virtualWanId"`
```
komalali, line 72: Looks like these String -> StringPtr changes might be breaking.

stack72: There's nothing specific we can do when something has been made optional...

stack72: I believe these are the kinda "breaking" changes we need to just accept

komalali (approved):
Looks like there are some potentially breaking changes due to required fields being made optional. Not sure if there's anything to be done there though.

PR #797, opened 2020-12-21
Upgrade to v2.41.0 of the AzureRM Terraform Provider

stack72:
Fixes: #796


provider/resource.go
```diff
@@ -819,7 +819,8 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_eventhub_cluster": {Tok: azureResource(azureEventHub, "Cluster")},
 
 			// Eventgrid
-			"azurerm_eventgrid_system_topic": {Tok: azureResource(azureEventGrid, "getSystemTopic")},
```
stack72, line 822: this is an expected rename that's "breaking" - we had mislabelled a resource as a datasource

mikhailshilkov (approved):

PR #847, opened 2021-03-01
Upgrade to v2.49.0 of the AzureRM Terraform Provider

stack72:
Fixes: #845


sdk/dotnet/NetApp/Snapshot.cs
```diff
@@ -283,10 +277,7 @@ public sealed class SnapshotState : Pulumi.ResourceArgs
 
         [Input("tags")]
         private InputMap<string>? _tags;
-
-        /// <summary>
-        /// A mapping of tags to assign to the resource.
-        /// </summary>
+        [Obsolete(@"This property as been deprecated as the API no longer supports tags and will be removed in version 3.0 of the provider.")]
```
mikhailshilkov, line 280: Nit: `version 3.0 of the provider` is confusing for us. If there is a straightforward way to change it to `4.0`, we could consider that.

mikhailshilkov (approved):
LGTM

sdk/dotnet/NetApp/Snapshot.cs
stack72: This is just pulling in the deprecation from the TF schema by default...

PR #854, opened 2021-03-12
Upgrade to v2.51.0 of the AzureRM Terraform Provider

stack72:
Fixes: #853


mikhailshilkov (approved):
LGTM except .NET tests are broken (the same trick with NuGet needed?)

PR #863, opened 2021-03-26
Upgrade to v2.53.0 of the AzureRM Terraform Provider

stack72:
Fixes: #862


provider/resource.go
```diff
@@ -590,6 +590,8 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_custom_provider":                    {Tok: azureResource(azureCore, "CustomProvider")},
 			"azurerm_resource_provider_registration":     {Tok: azureResource(azureCore, "ResourceProviderRegistration")},
 			"azurerm_subscription":                       {Tok: azureResource(azureCore, "Subscription")},
+			"azurerm_tenant_template_deployment":         {Tok: azureResource(azureCore, "TenantTemplateDeployment")},
+			"azurerm_template_spec_version":              {Tok: azureResource(azureCore, "TemplateSpec")},
```
mikhailshilkov, line 594: Should it not be `TemplateSpecVersion`?

stack72: great catch! This should actually have been a datasource as well :/

mikhailshilkov (approved):

PR #911, opened 2021-06-22
Upgrade to v2.64.0 of the AzureRM Terraform Provider

stack72:
Fixes: #910


provider/resource.go
```diff
@@ -797,7 +799,8 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_data_lake_store_virtual_network_rule": {Tok: azureResource(azureDatalake, "StoreVirtualNetworkRule")},
 
 			// Data Protection
-			"azurerm_data_protection_backup_vault": {Tok: azureResource(azureDataProtection, "BackupVault")},
+			"azurerm_data_protection_backup_vault":             {Tok: azureResource(azureDataProtection, "BackupVault")},
+			"azurerm_data_protection_backup_policy_postgresql": {Tok: azureResource(azureDataProtection, "BackupPolicyPostgresql")},
```
mikhailshilkov, line 803: I think we use "PostgreSql" for the namespace, should we do the same here? Although we have "DatasetPostgresql" too :(

mikhailshilkov (approved):
LGTM with one question

provider/resource.go
stack72: We use `LinkedServicePostgresql` more than `PostgreSql` so I think we should keep to that - the namespace we need to think about more

PR #929, opened 2021-07-23
Upgrade to v2.69.0 of the AzureRM Terraform Provider

stack72:
Fixes: #928﻿


provider/resource.go
```diff
@@ -45,99 +45,100 @@ const (
...
-	azureSearch                = "Search"                // Search
-	azureSignalr               = "SignalR"               // SignalR
-	azureSiteRecovery          = "SiteRecovery"          // SiteRecovery
-	azureSQL                   = "Sql"                   // SQL
-	azureStack                 = "Stack"                 // Stack
-	azureStorage               = "Storage"               // Storage
-	azureStreamAnalytics       = "StreamAnalytics"       // StreamAnalytics
-	azureSynapse               = "Synapse"               // Synapse
-	azureWaf                   = "Waf"                   // WAF
+	activeDirectoryDomainService = "ActiveDirectoryDomainService" // ActiveDirectoryDomainService
```
mikhailshilkov, line 48: Is it not just "DomainServices" as in https://github.com/Azure/azure-rest-api-specs/tree/master/specification/domainservices/resource-manager ?

provider/resource.go
```diff
@@ -1688,6 +1703,7 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_bot_web_app":               {Tok: azureResource(azureBot, "WebApp")},
 			"azurerm_bot_channel_ms_teams":      {Tok: azureResource(azureBot, "ChannelTeams")},
 			"azurerm_bot_channel_directline":    {Tok: azureResource(azureBot, "ChannelDirectLine")},
+			"azurerm_bot_channel_web_chat":      {Tok: azureResource(azureBot, "WebChat")},
```
mikhailshilkov, line 1706: Should it be `ChannelWebChat`?

stack72: Great catch! thanks @mikhailshilkov

provider/resource.go
mikhailshilkov: I think plural `domainservices` would be a better match than `domainservice`

mikhailshilkov (approved):
LGTM

Issue #945, opened 2021-09-14
Depend on node-fetch v2


roderik:
## Steps to reproduce

1. install node-fetch manually. Which gets you v3 since the beginning of September
2. Run pulumi up

This will fail with

```
Error [ERR_REQUIRE_ESM]: require() of ES Module <…>/node_modules/node-fetch/src/index.js from <…>/node_modules/@pulumi/azure/eventgrid/zMixins.js not supported.
```

Even though the TS code is correct according to the node-fetch changelog https://github.com/pulumi/pulumi-azure/blob/master/sdk/nodejs/eventgrid/zMixins.ts#L18

The compiled version is `const node_fetch_1 = require("node-fetch");`

I would assume that building the package would do this correctly, but alas.

Other pulumi packages depend on node-fetch directly, but azure does not 

e.g. kubernetes `"node-fetch": "^2.3.0",` which installs a v2 for this package. 

In any case, depending on a package used is a good practice, so doing that would fix this.

PR #946, opened 2021-09-14
fix: depend on node-fetch as it is used in the zMixins file and breaks when manually installing v3

roderik:
See #945


stack72:
Hey @roderik 

Thanks so much for this - we need to handle this via provider/resources.go and I included in it #947 

I will close this out as this is now going to be in that PR

Paul

roderik:
Ah, did not know that file existed, thanks!

PR #970, opened 2021-10-29
Upgrade to v2.83.0 of the AzureRM Terraform Provider

stack72:
Fixes: #969﻿

Please note there are a few non user facing changes here:

`azure.eventgrid.getSystemTopic` was originally a resource that was renamed to a data source, we now have a datasource for this so we removed the originally broken resource - that was from was a long time ago - maybe 2 major versions!

we had to override the MaxItems on ApplicationGateway so that we accept a list of ssl_policies still so that it is not a breaking change for our users


sdk/dotnet/AppPlatform/SpringCloudJavaDeployment.cs
```diff
@@ -94,18 +97,21 @@ public partial class SpringCloudJavaDeployment : Pulumi.CustomResource
         [Output("jvmOptions")]
         public Output<string?> JvmOptions { get; private set; } = null!;
 
-        /// <summary>
-        /// Specifies the required memory size of the Spring Cloud Deployment. Possible Values are between `1` and `8`. Defaults to `1` if not specified.
-        /// </summary>
         [Output("memoryInGb")]
-        public Output<int?> MemoryInGb { get; private set; } = null!;
+        public Output<int> MemoryInGb { get; private set; } = null!;
```
jkodroff, line 101: Breaking change.

sdk/dotnet/Network/Inputs/FirewallApplicationRuleCollectionRuleProtocolGetArgs.cs
```diff
@@ -15,8 +15,8 @@ public sealed class FirewallApplicationRuleCollectionRuleProtocolGetArgs : Pulum
         /// <summary>
         /// Specify a port for the connection.
         /// </summary>
-        [Input("port")]
-        public Input<int>? Port { get; set; }
+        [Input("port", required: true)]
+        public Input<int> Port { get; set; } = null!;
```
jkodroff, line 19: Breaking change, but it apparently became required upstream, so not much we can do about it.

jkodroff (approved):
All breaking changes are either in accordance with our policies or intentional per @stack72's comment above.

PR #980, opened 2021-11-19
Upgrade to v2.86.0 of the Azure RM Terraform Provider.

jkodroff:

PR #981, opened 2021-11-22
Ensure no regression to compute.OrchestratedVirtualMachineScaleSet zones

stack72:
PR #980 changed azure.compute.OrchestratedVirtualMachineScaleSet to have no MaxItems on zones parameter

This PR changes that to ensure we don't introduce a breaking change for our users. ﻿


mikhailshilkov (approved):
LGTM

PR #1025, opened 2022-02-04
Upgrade to v2.95.0 of the AzureRM Terraform Provider

stack72:
Fixes: #1024﻿


jkodroff (approved):
Per Slack convo with @stack72, AKA "Slack with Stack", the removed beta resources do not generate in the Pulumi provider due to that option being turned off in the bridge.

PR #1059, opened 2022-04-04
Upgrade to v3.0.2 of the AzureRM Terraform Provider

stack72:
Fixes: #1054
Fixes: #1055﻿


.github/workflows/command-dispatch.yml
```diff
@@ -14,7 +14,6 @@ env:
   PULUMI_API: https://api.pulumi-staging.io
   PULUMI_GO_DEP_ROOT: ${{ github.workspace }}/..
   PULUMI_LOCAL_NUGET: ${{ github.workspace }}/nuget
-  PULUMI_MISSING_DOCS_ERROR: true
```
jkodroff, line 17: How come we're getting rid of this? I set this to `true` after resolving all missing docs.

provider/resource.go
```diff
@@ -1880,10 +1800,12 @@ func Provider() tfbridge.ProviderInfo {
 
 			// Dashboard
 			// TODO: This resource is in the "Portal" module in the upstream provider. Move to a new namespace when terraform-provider-azurerm hits v3.0.
-			"azurerm_dashboard": {Tok: azureResource(azureDashboard, "Dashboard")},
+			"azurerm_dashboard":        {Tok: azureResource(azureDashboard, "Dashboard")},
```
jkodroff, line 1803: We should either move to the Portal module to be consistent with the upstream provider or remove the comment on the line above. If there's a reference for how we organize our modules, might be good to add to the README.

provider/resource.go
```diff
@@ -2092,10 +2014,14 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_management_group_template_deployment":      {Tok: azureResource(azureManagement, "GroupTemplateDeployment")},
 			"azurerm_management_group_subscription_association": {Tok: azureResource(azureManagement, "GroupSubscriptionAssociation")},
 			"azurerm_management_group_policy_assignment":        {Tok: azureResource(azureManagement, "GroupPolicyAssignment")},
+			"azurerm_management_group_policy_remediation":       {Tok: azureResource(azureManagement, "GroupPolicyRemediation")},
 
 			// communication
 			"azurerm_communication_service": {Tok: azureResource(azureCommunication, "Service")},
 
+			// connections
+			"azurerm_api_connection": {Tok: azureResource(azureConnections, "Connection")},
```
jkodroff, line 2023: I think this should be `ApiConnection`? If intentional, might be good to add a comment why it's not `ApiConnection`.

.github/workflows/command-dispatch.yml
stack72: Some of the new resources upstream have no docs I'm afraid :/

provider/resource.go
stack72: Tell me what you want me to do here and we will do it :)

.github/workflows/command-dispatch.yml
jkodroff: Let's add something like this (either the literal code or a func that returns the same) and keep this flag on so that we normalize missing docs as an error to be fixed: https://github.com/pulumi/pulumi-terraform-bridge/pull/494

provider/resource.go
jkodroff: Per Slack convo, the design intent is to mirror the upstream provider, so let's move to a Portal module.

.github/workflows/command-dispatch.yml
stack72: Fixed

jkodroff (approved):

PR #1104, opened 2022-07-06
[internal] Adding support for Java SDK to the pulumi-azure repo

stack72:


guineveresaenger (approved):
Same var naming concerns as in https://github.com/pulumi/pulumi-aws/pull/2040 but otherwise LGTM

PR #1125, opened 2022-08-19
Upgrade to v3.19.0 of the AzureRM Terraform Provider

stack72:
Fixes: #1124


jkisk (approved):

tmeckel:
@stack72 did you notice that Hashicorp just released a 3.19.1 version of the provider because they introduced another huge bug. This time with the Public DNS Zone and the corresponding records.

stack72:
@tmeckel I will get an update cut before release

tmeckel:
@stack72 Awesome! But TBH currently I wouldn't call Version 3 of the TF Azure provider some kind of stable software. 👎🏼 So many errors and bugs have been introduced. 😞

PR #1130, opened 2022-08-26
Add preConfigureCallback using Azure auth builder helper library

guineveresaenger:
Fixes https://github.com/pulumi/home/issues/2274

**UPDATE** I decided the third party library is useful and azure-focused enough that we should probably just surface its errors. I added that functionality.

Also I rebased off of default, so the tests should clear.

 Here's the output on a pulumi-azure program when not logged in:
```
@guin:azure-classic-go🦉 PATH=/Users/guin/go/src/github.com/pulumi/pulumi-azure/bin:$PATH pulumi up
Please choose a stack, or create a new one: dev
Previewing update (dev)

View Live: https://app.pulumi.com/guinevere/azure-classic-go/dev/previews/929a8af5-2459-420f-825c-cbc3f5efece3

     Type                         Name                  Plan       Info
 +   pulumi:pulumi:Stack          azure-classic-go-dev  create     
     └─ azure:core:ResourceGroup  exampleResourceGroup             1 error
 
Diagnostics:
  azure:core:ResourceGroup (exampleResourceGroup):
    error: failed to load application credentials:
    Details: obtain subscription() from Azure CLI: parsing json result from the Azure CLI: waiting for the Azure CLI: exit status 1: ERROR: Please run 'az login' to setup account.
    
    	Please make sure you have signed in via 'az login' or configured another authentication method.
    	See https://www.pulumi.com/registry/packages/azure/installation-configuration/ for more information.
```

### DIFF
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -285,9 +285,10 @@ func preConfigureCallback(vars resource.PropertyMap, c tfshim.ResourceConfig) er
 	_, err := builder.Build()
 
 	if err != nil {
-		return fmt.Errorf("failed to load application credentials.\n" +
-			"\tPlease sign in via 'az login' or configure another authentication method.\n" +
-			"\tSee https://www.pulumi.com/registry/packages/azure/installation-configuration/ for details.")
+		return fmt.Errorf("failed to load application credentials:\n"+
+			"Details: %v\n\n"+
+			"\tPlease make sure you have signed in via 'az login' or configured another authentication method.\n"+
+			"\tSee https://www.pulumi.com/registry/packages/azure/installation-configuration/ for more information.", err)
 	}
 	return nil
 }
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -286,7 +286,7 @@ func preConfigureCallback(vars resource.PropertyMap, c tfshim.ResourceConfig) er
 
 	if err != nil {
 		return fmt.Errorf("failed to load application credentials.\n" +
-			"\tPlease sign in via 'az login' or confugure another authentication method.\n" +
+			"\tPlease sign in via 'az login' or configure another authentication method.\n" +
 			"\tSee https://www.pulumi.com/registry/packages/azure/installation-configuration/ for details.")
 	}
 	return nil
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -271,7 +271,6 @@ func boolValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []stri
 func arrayValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) []string {
 	val, ok := vars[prop]
 	var vals []string
-	fmt.Println(vals)
 	if ok && val.IsArray() {
 		for _, v := range val.ArrayValue() {
 			vals = append(vals, v.StringValue())
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -252,6 +252,42 @@ func stringValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []st
 	return ""
 }
 
+// boolValue takes a bool value from a property map, then from environment vars; defaults to false
+func boolValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) bool {
+	val, ok := vars[prop]
+	if ok && val.IsBool() {
+		return val.BoolValue()
+	}
+	for _, env := range envs {
+		val, ok := os.LookupEnv(env)
+		if ok && val == "true" {
+			return true
+		}
+	}
+	return false
+}
+
+// arrayValue takes an array value from a property map, then from environment vars; defaults to an empty array
+func arrayValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) []string {
+	val, ok := vars[prop]
+	var vals []string
+	fmt.Println(vals)
+	if ok && val.IsArray() {
+		for _, v := range val.ArrayValue() {
+			vals = append(vals, v.StringValue())
+		}
+		return vals
+	}
+
+	for _, env := range envs {
+		val, ok := os.LookupEnv(env)
+		if ok {
+			return strings.Split(val, ";")
+		}
+	}
+	return vals
+}
+
 // preConfigureCallback returns an error when cloud provider setup is misconfigured
 func preConfigureCallback(vars resource.PropertyMap, c tfshim.ResourceConfig) error {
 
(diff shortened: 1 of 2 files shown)


provider/resource.go
```diff
@@ -235,6 +237,60 @@ func detectCloudShell() cloudShellProfile {
...
+func preConfigureCallback(vars resource.PropertyMap, c tfshim.ResourceConfig) error {
+
+	envName := stringValue(vars, "environment", []string{"ARM_ENVIRONMENT"})
+	if envName == "" {
+		envName = "public"
+	}
+
+	// validate the azure config
+	// make a Builder
+	builder := &authentication.Builder{
```
thomas11, line 265: Ideally, we would find a way to share this builder code with `provider.go:getAuthConfig`. I assume the main issue is that we don't have the provider config yet for `k.getConfig`? We could abstract the lookup of a single property into a `configGetter` interface and pass it in, for instance.

provider/resource.go
```diff
@@ -235,6 +237,60 @@ func detectCloudShell() cloudShellProfile {
...
+		SupportsClientSecretAuth: true,
+		//SupportsManagedServiceIdentity: useMsi,
+		SupportsAzureCliToken: true,
+		//SupportsAuxiliaryTenants:       len(auxTenants) > 0,
+	}
+
+	_, err := builder.Build()
+
+	if err != nil {
+		return fmt.Errorf("failed to load application credentials. Please run 'az login' to set up account.\n" +
```
thomas11, line 288: I suggest wording it a little differently. In Azure, you authenticate as an application *or* a user. `az login` is only correct for the latter. Maybe

> failed to authenticate to Azure. Please sign in via `az login` or configure another authentication method. See...

provider/resource.go
```diff
@@ -235,6 +237,60 @@ func detectCloudShell() cloudShellProfile {
 	return negative
 }
 
+// stringValue gets a string value from a property map if present, else ""
```
thomas11, line 240: Comment doesn't match the func because of `envs`

provider/resource.go
guineveresaenger: I'm not sure I understand your comment entirely - we currently [do not have CheckConfig enabled](https://github.com/pulumi/pulumi-terraform-bridge/blob/master/pkg/tfbridge/provider.go#L314) in the TF bridge, and would have to implement such an interface there so we can capture an auth config getter for any cloud? 
here's the issue: https://github.com/pulumi/pulumi-terraform-bridge/issues/244

thomas11: Oops, I was confused and was talking about provider.go in the native azure provider. I has a very similar func, but I guess sharing code is not practical.

provider/resource.go
```diff
@@ -235,6 +237,61 @@ func detectCloudShell() cloudShellProfile {
...
+		//SupportsManagedServiceIdentity: useMsi,
+		SupportsAzureCliToken: true,
+		//SupportsAuxiliaryTenants:       len(auxTenants) > 0,
+	}
+
+	_, err := builder.Build()
+
+	if err != nil {
+		return fmt.Errorf("failed to load application credentials.\n" +
+			"\tPlease sign in via 'az login' or confugure another authentication method.\n" +
```
jkisk, line 289: confugure > configure

provider/resource.go
```diff
@@ -235,6 +237,61 @@ func detectCloudShell() cloudShellProfile {
...
+			"\tSee https://www.pulumi.com/registry/packages/azure/installation-configuration/ for details.")
+	}
```
jkisk, line 291: I think it might be useful to pass along the original error as well, but I understand that might not be desirable.

jkisk (commented):
Would you be able to post a before and after image in PR? always nice for UI changes.

provider/resource.go
guineveresaenger: the original error is not too bad, to be honest - but we're trying to avoid accidentally passing along TF-specific messages. I've followed the pattern from the other major cloud providers by overwriting. I feel a bit meh about it too but came down on the overwrite side. :)

jkisk: That makes sense, especially in these providers.

jkisk (approved):
Looks good, once we can get the build issue sorted and tests are clear.  Thank you, this error is much more helpful!

provider/resource.go
```diff
@@ -235,6 +237,62 @@ func detectCloudShell() cloudShellProfile {
...
+	builder := &authentication.Builder{
+		SubscriptionID:     stringValue(vars, "subscriptionID", []string{"ARM_SUBSCRIPTION_ID"}),
+		ClientID:           stringValue(vars, "clientId", []string{"ARM_CLIENT_ID"}),
+		ClientSecret:       stringValue(vars, "clientSecret", []string{"ARM_CLIENT_SECRET"}),
+		TenantID:           stringValue(vars, "tenantId", []string{"ARM_TENANT_ID"}),
+		Environment:        envName,
+		ClientCertPath:     stringValue(vars, "clientCertificatePath", []string{"ARM_CLIENT_CERTIFICATE_PATH"}),
+		ClientCertPassword: stringValue(vars, "clientCertificatePassword", []string{"ARM_CLIENT_CERTIFICATE_PASSWORD"}),
+		MsiEndpoint:        stringValue(vars, "msiEndpoint", []string{"ARM_MSI_ENDPOINT"}),
+		//AuxiliaryTenantIDs:   auxTenants,
```
thomas11, line 274: What's the reason for the commented lines? Should we uncomment or remove them?

guineveresaenger: Thank you for catching the commented-out settings - this was an oversight on my part, as they take a little extra implementation, not being underlying string values.
[The upstream docs are a bit minimal here](https://registry.terraform.io/providers/hashicorp/azurerm/latest/docs#auxiliary_tenant_ids) 
I [confirmed that the separator expected in the command line is a `;`](https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/provider/provider.go#L328)

provider/resource.go
```diff
@@ -252,6 +252,42 @@ func stringValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []st
...
+		for _, v := range val.ArrayValue() {
+			vals = append(vals, v.StringValue())
+		}
+		return vals
+	}
+
+	for _, env := range envs {
+		val, ok := os.LookupEnv(env)
+		if ok {
+			return strings.Split(val, ";")
```
thomas11, line 285: do we need to handle "; " with a space?

provider/resource.go
```diff
@@ -252,6 +252,42 @@ func stringValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []st
...
+		}
+	}
+	return false
+}
+
+// arrayValue takes an array value from a property map, then from environment vars; defaults to an empty array
+func arrayValue(vars resource.PropertyMap, prop resource.PropertyKey, envs []string) []string {
+	val, ok := vars[prop]
+	var vals []string
+	fmt.Println(vals)
```
thomas11, line 274: remove?

provider/resource.go
guineveresaenger: I really don't think so: https://github.com/hashicorp/terraform-provider-azurerm/blob/main/internal/provider/provider.go#L328

thomas11 (approved):

PR #1152, opened 2022-10-20
Upgrade terraform provider to 3.27.0

danielrbradley:
This fixes an incorrect name for azurerm_cdn_frontdoor_custom_domain and add alias to correct mistake `FrontdoorCustomDomain` was incorrectly named `FrontdoorCustomEndpoint` in `v5.21.0`. I've added a type alias so it should be a simple rename which will not cause any deletion or recreation of resources.


provider/resource.go
```diff
@@ -950,7 +955,8 @@ func Provider() tfbridge.ProviderInfo {
 					}),
 				},
 			},
-			"azurerm_cdn_frontdoor_custom_domain":                        {Tok: azureResource(azureCDN, "FrontdoorCustomEndpoint")},
+			"azurerm_cdn_frontdoor_custom_domain":                        {Tok: azureResource(azureCDN, "FrontdoorCustomDomain"), Aliases: []tfbridge.AliasInfo{{Type: strRef("azure:cdn/frontdoorCustomEndpoint:FrontdoorCustomEndpoint")}}},
```
thomas11, line 958: A comment would be nice here to help future readers understand the alias?

thomas11 (approved):

viveklak (approved):
Reminder to changelog the alias introduction.

PR #1165, opened 2022-11-21
[provider] Fix OIDC support

pgavlin:
Add parameters to the client configuration used by the pre-configure callback so that the provider can support OIDC authentication.


thomas11:
Superseded by #1245

Issue #1190, opened 2023-01-12
Go SDK does not compile 


t0yv0:
### What happened?

Go code on master does not compile.

### Steps to reproduce

git rev-parse HEAD                                                                                                                                                          ~/code/pulumi-azure/sdk/go
083163aeaa4d3cddc21cfe528ad8f0d5c0e3a2bc

cd sdk/go && go build -v all

```
azure/cdn/frontdoorCustomEndpoint.go:125:6: undefined: FrontdoorCustomEndpointTlsOutput
azure/cdn/frontdoorCustomEndpoint.go:143:17: invalid operation: args.Tls == nil (operator == not defined on untyped nil)
azure/cdn/frontdoorCustomEndpoint.go:181:7: undefined: FrontdoorCustomEndpointTls
azure/cdn/frontdoorCustomEndpoint.go:200:6: undefined: FrontdoorCustomEndpointTlsPtrInput
azure/cdn/frontdoorCustomEndpoint.go:221:6: undefined: FrontdoorCustomEndpointTls
azure/cdn/frontdoorCustomEndpoint.go:237:6: undefined: FrontdoorCustomEndpointTlsInput
azure/cdn/frontdoorCustomEndpoint.go:358:46: undefined: FrontdoorCustomEndpointTlsOutput
azure/cdn/frontdoorCustomEndpoint.go:359:51: undefined: FrontdoorCustomEndpointTlsOutput
```

### Expected Behavior

Generated GO SDK compiles

### Actual Behavior

Compilation failure.

### Output of `pulumi about`

NA

### Additional context

This blocks updating to latest ci-mgmt which starts checking for the Go code to be compile-able.

It might be an instance of Pulumi codegen bug:

https://github.com/pulumi/pulumi/issues/11837

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

Issue #1160, opened 2022-11-09
`make clean` is too aggressive for nodejs and dotnet


thomas11:
### What happened?

`make clean build` on a clean master branch fails with

```
error: failed to generate package: open pulumi-azure/sdk/nodejs/location.ts: no such file or directory
```

Building the SDKs separately shows that .Net is also affected:
```
error: failed to generate package: open pulumi-azure/sdk/dotnet/Storage/SharedAccessSignature.cs: no such file or directory
```

Go and Python are not affected.

History shows that the seemingly missing files have a history of manual changes:
```
git lg -- sdk/nodejs/location.ts
* 4df16916e - Add the UAE Azure regions that were opened (3 years, 5 months ago) <stack72>
* 81ba19b11 - Export Locations module for correct Location namespacing (3 years, 5 months ago) <stack72>
* a870d88e8 - Adding a mixin for Azure Locations (3 years, 5 months ago) <stack72>
```

### Steps to reproduce

`make clean build`

### Expected Behavior

SDKs are rebuilt.

### Actual Behavior

see above?

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

PR #1192, opened 2023-01-13
Re-generate Go SDK

t0yv0:
Fixes https://github.com/pulumi/pulumi-azure/issues/1190

This makes the Go SDK compile again.

```
rm -rf sdk/go
make provider
make build_go
```

There is an issue with makefiles not deleting the prior SDK before re-generating. Causing left-over files in the source tree.


thomas11:
#1160 is the reason we don't `clean` automatically before building the SDKs, FYI @t0yv0

t0yv0:
Eek, unfortunate! We could ask for a codegen feature to maintain a list of files it generates, then delete only those. Platform team has something like this in the test repos for generated code.

PR #1199, opened 2023-01-24
Upgrade terraform-provider-azurerm to v3.40.0

thomas11:
- Upgrade terraform-provider-azurerm to v3.40.0
- Rebuild SDKs for v3.40.0


danielrbradley (approved):
✅  Resource mapping names look good
✅  Docs changes look good
❓  Looks like upstream just typo'd the 'blob' as 'blog' in the docs. Expect this will get fixed in the next release and we'll have to watch out for this during the next review and remove the override?

thomas11:
> Looks like upstream just typo'd the 'blob' as 'blog' in the docs. Expect this will get fixed in the next release and we'll have to watch out for this during the next review and remove the override?

@danielrbradley I'm hoping our tooling will error if they rename the file, just like it did before I added the explicit "Docs" reference.

Issue #1201, opened 2023-01-30
Azure Function App v4 - no support to specify bundle version


imbelchev:
https://github.com/pulumi/pulumi-azure/blob/151eba6e19cc41e70cbc5ffb1817cb777649ead9/sdk/nodejs/appservice/zMixins.ts#L397

With Azure Function App version `~4` the extension bundle version needs to be updated accordingly. Currently, no support to specify the extension bundle version.

https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#extension-bundles 

Maybe I am missing how to configure it?

thomas11:
Hi @imbelchev, thanks for raising this. It doesn't seem that it can be configured currently.

I'm not a Functions expert. Would it be enough to change the version range to `[1.*, 4.*]` or similar? Or do different users need different ranges?

imbelchev:
Hi @thomas11,

Thanks for the fast replay.

I am not really a big expert on Azure Functions as well but, from how I understand things - if `[1.*, 4.*]` it should automatically pick up the correct version.

Supported major versions of Azure Functions:
* Functions v1 is still in support.
* Functions v2 is no longer supported.
* Functions v3 end of support is 13-Dec-2022, so probably soon will not be supported.
* Functions v4 is the recommended one

imbelchev:
So I have tested this. There is a mapping `runtime version : extension bundle version` provided by Microsoft which needs to be strictly followed. It's from the same link posted in the issue initially - [Register Azure Functions binding extensions](https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#extension-bundles)

![image](https://user-images.githubusercontent.com/61695701/215868105-a5502807-7fb1-447e-80c3-17e17018f0a8.png)

imbelchev:
@thomas11 ,

Are these changes enough to accept the PR and release a new version?
https://github.com/pulumi/pulumi-azure/pull/1202

thomas11:
> @thomas11 , @mikhailshilkov , As the contribution was accepted and PR merged, do you have any timeline for when this change might get to an actual public release?

@imbelchev  It should be out by the end of the week unless something unexpected happens.

imbelchev:
@thomas11 , awesome ❤️ 

Thank you!

danielrbradley:
Fixed by https://github.com/pulumi/pulumi-azure/pull/1202

PR #1202, opened 2023-01-31
Azure Function App v4 support

imbelchev:
Issue:[Azure Function App v4 - no support to specify bundle version](https://github.com/pulumi/pulumi-azure/issues/1201)

### DIFF
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -388,6 +388,10 @@ interface DeploymentArchiveResult {
 }
 
 async function produceDeploymentArchiveAsync(args: MultiCallbackFunctionAppArgs): Promise<DeploymentArchiveResult> {
+
+    // Different runtime versions now require different extension bundle version ranges
+    // This mapping is described in the following MS doc
+    // https://learn.microsoft.com/en-us/azure/azure-functions/functions-bindings-register#extension-bundles
     let extentionBundleVersion = new Map([
         ['~1', '[1.*, 2.0.0)'],
         ['~2', '[2.*, 3.0.0)'],
--- a/sdk/nodejs/appservice/zMixins.ts
+++ b/sdk/nodejs/appservice/zMixins.ts
@@ -176,7 +176,7 @@ interface FunctionAppArgsBase {
 
     /**
      * Controls the value of WEBSITE_NODE_DEFAULT_VERSION in `appSettings`.  If not provided,
-     * defaults to `~12`.
+     * defaults to `~14`.
      */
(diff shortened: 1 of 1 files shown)


sdk/nodejs/appservice/zMixins.ts
```diff
@@ -394,7 +394,7 @@ async function produceDeploymentArchiveAsync(args: MultiCallbackFunctionAppArgs)
         tracing: { consoleLevel: "verbose" },
         extensionBundle: {
             id: "Microsoft.Azure.Functions.ExtensionBundle",
-            version: "[1.*, 2.0.0)"
+            version: "[3.3.0, 4.0.0)"
```
thomas11, line 397: As you pointed out on #1201, there's an enforced mapping between extension bundle and runtime version. If we change this, won't we break users that are currently using pulumi-azure with 1.x functions?

imbelchev: @thomas11,
How about now? Just a simple map that should get it going for all versions?

imbelchev: Thanks Thomas,

Please find answers to your points below:

* "I see the previous default version was ~3 but the bundle version was [1...]. How did that work if the mapping is strict, shouldn't ~3 have been matched with [3.3....]? Or did the default path just never work?"
> I remember reading somewhere from the docs, that for the previous versions of Function Apps the extension bundle version was not required to be specified. However, this was changed whit the release of v4. 

* "Can you please put a comment at the mapping table explaining why we need this, and maybe a link to the MS docs?"
> Sure, I can do that.

* "How can we test this? Did you test it on your end with a ~4 function?"
> I have tested this with v3 and v4 functions - working as expected.

* "I wonder if we should err on the safe side and default to ~1 (or the previous ~3), to avoid breaking existing users."
> I am not sure if this is going to break something. If users are following best practices they should check the preview carefully before actually applying any change. I can confirm that the diff in the preview is clear enough to sign that the version is being changed. They can either act accordingly and set version ~3 or consider migrating to ~4. Especially when v3 is EOL now and no service support is being provided.
The current default version value was set to ~3, so if there are users out there that are using v1 - then they have probably already specified ``version: "~1"``

imbelchev: @thomas11, Is there anything else we need to discuss?

thomas11: I think we're all good! Thank you so much for the contribution!

imbelchev: Thank you @thomas11

PR #1210, opened 2023-02-14
Update terraform-provider-azurerm to v3.43.0

rquitales:


sdk/dotnet/AppService/ActiveSlot.cs
```diff
@@ -9,42 +9,6 @@
 
 namespace Pulumi.Azure.AppService
 {
-    /// <summary>
```
thomas11, line 12: Looks like we're losing lots of examples again :( We should investigate this before releasing.

rquitales: And CI just failed on a dirty work dir...

rquitales: Seems like a local env issue with generating. Regenerated and pushed changes.

thomas11 (approved):

PR #1233, opened 2023-03-14
Fix useMsi Authorization Config

brucetan-displayr:
https://www.pulumi.com/registry/packages/azure/installation-configuration/#configuration-options

According to azure classic documentation, `useMsi` should match the environment variable `ARM_USE_MSI`

image


thomas11 (approved):
thank you!

Issue #1036, opened 2022-02-18
Eventgrid test broken


jkodroff:
The acceptance test in `examples/eventgrid/` is (at the time of writing), broken in both the `master.yml` and `cron.yaml` flows:

* master: https://github.com/pulumi/pulumi-azure/runs/5250503578?check_suite_focus=true
* cron: https://github.com/pulumi/pulumi-azure/runs/5250566690?check_suite_focus=true

They both have the same failure:

```
  Diagnostics:
    pulumi:pulumi:Stack (azure-eventgrid-p-it-fv-az201-2-eventgrid-fb02cdf7):
      error: Running program '/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689' failed with an unhandled exception:
      TSError: ⨯ Unable to compile TypeScript:
      index.ts(11,44): error TS2339: Property 'subject' does not exist on type 'ResourceGroupEvent'.
        Property 'subject' does not exist on type 'ResourceActionCancelEventData'.
      index.ts(12,47): error TS2339: Property 'eventType' does not exist on type 'ResourceGroupEvent'.
        Property 'eventType' does not exist on type 'ResourceActionCancelEventData'.
      index.ts(33,44): error TS2339: Property 'subject' does not exist on type 'StorageBlobCreatedEventData'.
      index.ts(49,44): error TS2339: Property 'subject' does not exist on type 'StorageBlobDeletedEventData'.
      
          at createTSError (/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689/node_modules/ts-node/src/index.ts:261:12)
          at getOutput (/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689/node_modules/ts-node/src/index.ts:367:40)
          at Object.compile (/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689/node_modules/ts-node/src/index.ts:558:11)
          at Module.m._compile (/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689/node_modules/ts-node/src/index.ts:439:43)
          at Module._extensions..js (internal/modules/cjs/loader.js:1114:10)
          at Object.require.extensions.<computed> [as .ts] (/tmp/p-it-fv-az201-2-eventgrid-fb02cdf7-577505689/node_modules/ts-node/src/index.ts:442:12)
          at Module.load (internal/modules/cjs/loader.js:950:32)
[...]

PR #1247, opened 2023-03-31
Re-enable eventgrid test

mikhailshilkov:
Check if maybe it passes now?

Fix https://github.com/pulumi/pulumi-azure/issues/1036


mikhailshilkov:
Still failed... this needs another good look

PR #1250, opened 2023-04-12
Update to v3.51.0

kpitzen:


sdk/java/src/main/java/com/pulumi/azure/automation/Schedule.java
```diff
@@ -223,14 +223,14 @@ public Output<String> startTime() {
         return this.startTime;
     }
     /**
-     * The timezone of the start time. Defaults to `UTC`. For possible values see: &lt;https://docs.microsoft.com/en-us/rest/api/maps/timezone/gettimezoneenumwindows&gt;
+     * The timezone of the start time. Defaults to `Etc/UTC`. For possible values see: &lt;https://docs.microsoft.com/en-us/rest/api/maps/timezone/gettimezoneenumwindows&gt;
```
thomas11, line 226: There are many instances of this change. Looks weird, doesn't it? What's `Etc/UTC`?

kpitzen: Seems like it's just conforming to the `location/timezone` standard here. I think it's fine?

thomas11: Ok, this is me not knowing the relevant standards apparently :)

thomas11 (approved):

PR #1252, opened 2023-04-20
Upgrade terraform-provider-azurerm to v3.52.0

mikhailshilkov:


kpitzen (approved):

kpitzen:
Approved since no breaking schema changes, but do fields being marked as secret ever cause breaks?

mikhailshilkov:
> do fields being marked as secret ever cause breaks?

I don't think so? At least, I don't remember hearing about this problem earlier. Will go ahead with the merge.

PR #1260, opened 2023-05-02
Default Token Mapping

### LINKED ISSUE #1254: Add automatic token mapping
## Hello!

- Vote on this issue by adding a 👍 reaction
- If you want to implement this feature, comment to let us know (we'll work with you on design, scheduling, etc.)

## Issue details

We should apply automatic token mapping to minimize the manual intervention necessary to upgrade a provider.

### Affected area/feature

Token mapping is applied in `resources.go`.

iwahbe:
Fixes #1254 

Apply default token mappings.

Certain fields were removed to test that auto-mapping worked.

### DIFF
--- a/provider/resource.go
+++ b/provider/resource.go
@@ -178,8 +178,11 @@ var moduleMap = map[string]string{
 	"api_management":       azureAPIManagement,
 	"app":                  azureAppConfiguration,
 	"application_insights": azureAppInsights,
-	// Ignored: azureAppPlatform. Resources and datasources with this prefix are mapped into
+
+	// Resources and datasources with this prefix are mapped from
 	// azurerm_spring_${rest} to ${pkg}:AppPlatform:camel(spring_${rest}).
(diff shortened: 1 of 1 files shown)


provider/resource.go
```diff
@@ -32,6 +32,7 @@ import (
 	"github.com/hashicorp/terraform-provider-azurerm/shim"
 	"github.com/pulumi/pulumi-azure/provider/v5/pkg/version"
 	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge"
+	"github.com/pulumi/pulumi-terraform-bridge/v3/pkg/tfbridge/x"
```
thomas11, line 35: x?

thomas11 (approved):

provider/resource.go
iwahbe: The API is considered "experimental", and `/x` is the standard go way to indicate that. We have this deployed in most of our providers, so this is more about out-of-company api flexibility at this point.

PR #1284, opened 2023-06-02
Update master.yml

guineveresaenger:
Use updated build flag and goreleaser action


AaronFriel (approved):

t0yv0 (changes requested):
Keeping up to date is good. Note though that this is auto-gen file so ideally these changes should be coming from ci-mgmt change. Perhaps if you could make this change in ci-mgmt and we merge it there it'd roll out to all providers without further action from us (auto merge). That'd be ideal! Thank you.

iwahbe:
@guineveresaenger @t0yv0 should this PR go forward, or should I close it?

PR #1289, opened 2023-06-06
Experiment with faster goreleaser

t0yv0:
Attempting to work around out-of-disk issues in goreleaser pass by removing the need to run make tfgen again.


t0yv0:
This did not quite work out, still hitting out of disk issues.

PR #1384, opened 2023-10-12
Run `make ci-mgmt` to regen GHA workflows

rquitales:
Faster than waiting for the nightly sync to regen and fix the broken CI on master branch.

### DIFF
--- a/.github/workflows/master.yml
+++ b/.github/workflows/master.yml
@@ -135,7 +135,7 @@ jobs:
       COVERAGE_OUTPUT_DIR: ${{ secrets.COVERAGE_OUTPUT_DIR }}
     name: generate_coverage_data
     needs: prerequisites
-    runs-on: pulumi-ubuntu-8core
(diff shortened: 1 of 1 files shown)


.github/workflows/master.yml
```diff
@@ -135,7 +135,7 @@ jobs:
       COVERAGE_OUTPUT_DIR: ${{ secrets.COVERAGE_OUTPUT_DIR }}
     name: generate_coverage_data
     needs: prerequisites
-    runs-on: pulumi-ubuntu-8core
+    runs-on: ubuntu-latest
```
t0yv0, line 138: This change reduces the runner size to a smaller machine. Is this what we want?

rquitales: Yes, `generate_coverage_data` was mistakenly given the larger runner instead of `prerequisites` for the master workflow.

t0yv0 (approved):

PR #1385, opened 2023-10-12
Increase default runner to pulumi-ubuntu-8core

rquitales:
To help unblock all CI jobs.

### DIFF
--- a/.ci-mgmt.yaml
+++ b/.ci-mgmt.yaml
@@ -28,6 +28,7 @@ team: ecosystem
 goBuildParallelism: 2
 javaGenVersion: "v0.8.0"
 runner:
+  default: pulumi-ubuntu-8core
   prerequisites: pulumi-ubuntu-8core
   buildSdk: pulumi-ubuntu-8core
   publish: pulumi-ubuntu-8core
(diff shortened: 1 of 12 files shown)


.ci-mgmt.yaml
```diff
@@ -28,6 +28,7 @@ team: ecosystem
 goBuildParallelism: 2
 javaGenVersion: "v0.8.0"
 runner:
+  default: pulumi-ubuntu-8core
   prerequisites: pulumi-ubuntu-8core
   buildSdk: pulumi-ubuntu-8core
   publish: pulumi-ubuntu-8core
```
iwahbe, line 34: Nit: We can comment these out if we want, since they are overriding the default *and* the same as the default.

iwahbe (approved):

guineveresaenger (changes requested):
I think most of these aren't really necessary.

Most of these were addressed via logic in the ci-mgmt templates (the templates decide which runner to assign based on the config field set), and passed to pulumi-azure's Workflows via [PRs like this one](https://github.com/pulumi/pulumi-azure/commit/f2928e097779e9a0f039665b26955faf60520032).

See also changes applied [here](https://github.com/pulumi/pulumi-azure/blob/master/.github/workflows/release.yml#L222).

I think we can close this out.

iwahbe:
@rquitales Can we close this?

rquitales:
Stale PR.

PR #1393, opened 2023-10-23
Upgrade pulumi-terraform-bridge to v3.62.0

danielrbradley:
This PR was generated via `$ upgrade-provider pulumi/pulumi-azure --upstream-provider-name=terraform-provider-azurerm --kind=all`.

---

- Upgrading pulumi-terraform-bridge from v3.61.0 to v3.62.0.


iwahbe (commented):
It look like you have an issue with your java versions.

t0yv0:
v3.62.0 brings in a known P1 from pulumi/pkg recommend waiting till the next version.

t0yv0:
https://github.com/pulumi/pulumi-terraform-bridge/issues/1451 is pinned in the bridge

I think we can fix forward now finally; the fixes are available in latest pulumi/pkg. Releasing new bridge shortly.

PR #1469, opened 2023-11-28
Upgrade terraform-provider-azurerm to v3.82.0

### LINKED ISSUE #1465: Upgrade terraform-provider-azurerm to v3.82.0
Release details: https://github.com/hashicorp/terraform-provider-azurerm/releases/tag/v3.82.0

guineveresaenger:
This PR was generated via `$ upgrade-provider pulumi/pulumi-azure`.

---

- Upgrading terraform-provider-azurerm from 3.81.0  to 3.82.0.
	Fixes #1465


patches/0005-Modify-resources.patch
```diff
@@ -1284,21 +1284,25 @@ index 5f00d7466f..0e7f4ae71b 100644
...
--						}, false),
+@@ -1426,8 +1428,10 @@ func resourceMsSqlDatabaseSchema() map[string]*pluginsdk.Schema {
+ 						Type:     pluginsdk.TypeString,
+ 						Optional: true,
+ 						Default:  EmailAccountAdminsStatusDisabled,
+-						ValidateFunc: validation.StringInSlice(PossibleValuesForEmailAccountAdminsStatus(),
+-							false),
++						ValidateFunc: validation.StringInSlice([]string{
++							"Disabled",
++							"Enabled",
```
iwahbe, line 1295: I'm having trouble reading this patch, but I think we are replacing `PossibleValuesForEmailAccountAdminsStatus(), false` (new upstream) with `[]string{"Disabled", "Enabled"}, features.CaseInsensitive()` and I'm not sure why. I believe that we used to be replacing `[]string{"Disabled", "Enabled"}, false` (old upstream) with `[]string{"Disabled", "Enabled"}, features.CaseInsensitive()`. Can we track upstream's change and replace with `PossibleValuesForEmailAccountAdminsStatus(), features.CaseInsensitive()`?

I have no issue believing that the answer to my question is "no", but I'd like to understand why not.

iwahbe (approved):
One question on a patch change, otherwise LGTM

patches/0005-Modify-resources.patch
guineveresaenger: ah! yes, the upstream changes are equivalent to our patch - `PossibleValuesForEmailAccountAdminsStatus()` contains only "Disabled" and "Enabled".

We should remove.

PR #1473, opened 2023-11-30
Fix ProgramTest step paths in tests

thomas11:
These paths were not found but the error was silently ignored, see https://github.com/pulumi/pulumi/issues/14694.


iwahbe (approved):

iwahbe:
One of the modified tests seems to be failing.

PR #1567, opened 2023-12-27
Do not assume ExpectRefreshChanges: true for all tests

t0yv0:
Before this change unexpected refresh change were ignored for all example-based integration tests via `ExpectRefreshChanges: true`, so accidentally regressing on this would have gone unnoticed. After the PR such settings are only set on examples that currently need them and they are cross-linked with issues causing the problem.

### DIFF
--- a/examples/examples_dotnet_test.go
+++ b/examples/examples_dotnet_test.go
@@ -31,6 +31,6 @@ func TestAccAppServiceCs(t *testing.T) {
 				assert.Equal(t, body, "Hello Pulumi")
 			}),
 		})
-
+	skipRefresh(&test)
 	integration.ProgramTest(t, &test)
 }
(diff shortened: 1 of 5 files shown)


t0yv0:
Found an interesting problem: https://github.com/pulumi/pulumi-azure/issues/1568

t0yv0:
Another set of failures are due to https://github.com/pulumi/pulumi-terraform-bridge/issues/1595

t0yv0:
LB example further cycles because of https://github.com/pulumi/pulumi-azure/issues/1572 - I fixed the example but leaving the issue in case there are more problems of this sort in general, as there likely are.

iwahbe (approved):
Should `protocol` be case insensitive?

Otherwise LGTM

t0yv0:
Yeah that's filed in #1572.

PR #1626, opened 2024-01-19
Move Upstream version tag to correct commit

### LINKED ISSUE #1624: Workflow failure: Check upstream upgrade
## Workflow Failure

[Check upstream upgrade](https://github.com/pulumi/pulumi-azure/blob/master/.github/workflows/check-upstream-upgrade.yml) has failed. See the list of failures below:

- [2024-01-19T03:05:18.000Z](https://github.com/pulumi/pulumi-azure/actions/runs/7578901412)

guineveresaenger:
The upstream tag for v3.87.0 was moved since the release.

This PR uses the updated upstream tag (same tag name, different commit) to fix upgrade automations.

Because there were a few minor updates between the OG v3.87.0 tag, we should release a patch to reflect those updates.

Fixes #1624.

- Checkout upstream at new v3.87.0 tag and go mod tidy
- gen schema and make sdks

### DIFF
--- a/provider/go.mod
+++ b/provider/go.mod
@@ -6,7 +6,7 @@ toolchain go1.21.3
 
 require (
 	github.com/Azure/go-autorest/autorest/azure/cli v0.4.6
-	github.com/hashicorp/go-azure-sdk v0.20240111.1094251
+	github.com/hashicorp/go-azure-sdk v0.20240112.1095456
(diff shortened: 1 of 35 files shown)


iwahbe (approved):
It looks like something is wrong with your environment, and you will need to re-run `make tfgen build_sdks`. Other than that, LGTM.

I agree on releasing this as a patch.

PR #1630, opened 2024-01-22
upgrade terraform provider azurerm to v3.88.0

### LINKED ISSUE #1627: Upgrade terraform-provider-azurerm to v3.88.0
Release details: https://github.com/hashicorp/terraform-provider-azurerm/releases/tag/v3.88.0

guineveresaenger:
This PR was partially generated via `$ upgrade-provider pulumi/pulumi-azure`.

---

- Upgrading terraform-provider-azurerm from 3.87.0  to 3.88.0.
	Fixes #1627 

- Add new Compute (VirtualMachine) resource
- make sdks


provider/resources.go
```diff
@@ -1058,6 +1058,7 @@ func Provider() tfbridge.ProviderInfo {
 			"azurerm_virtual_machine_scale_set":                {Tok: azureResource(azureCompute, "ScaleSet")},
 			"azurerm_virtual_machine_packet_capture":           {Tok: azureResource(azureCompute, "PacketCapture")},
 			"azurerm_virtual_machine_scale_set_packet_capture": {Tok: azureResource(azureCompute, "ScaleSetPacketCapture")},
+			"azurerm_virtual_machine_run_command":              {Tok: azureResource(azureCompute, "RunCommand")},
```
iwahbe, line 1061: I believe that all `azurerm_virtual_machine_*` resources map to `azureCompute`. Instead of manually mapping, can we add `virtual_machine` to the set of auto-mapped modules and remove lines 1057-1061+?

guineveresaenger: According to [this comment](https://github.com/pulumi/pulumi-azure/blob/7fdbb9e3b86b5620c9303bd639305680192d71a6/provider/resources.go#L217) a choice was made in the past to explicitly not map azureCompute.

iwahbe: My apologies. That comment was overly conservative, from when we had much less confidence in how mappings worked (and from when we didn't have good support for n:1 mappings).

We are now able to easily support this scenario, and we have a clear directive: all `azurerm_virtual_machine_*` commands are mapped into `azureCompute`.

guineveresaenger: I don't think this is correct.

if we map `	"virtual_machine": azureCompute,` then we get a whole new ComputeVirtualMachine module, e.g. sdk/nodejs/computevirtualmachine/.

Perhaps I'm misunderstanding something.

If this is a more involved change, then perhaps we should merge this PR as-is so as to avoid a very late release, and follow up separately.

iwahbe (approved):
LGTM

PR #1656, opened 2024-01-27
Upgrade pulumi-terraform-bridge to v3.72.0

guineveresaenger:
**NOTE FOR THE REVIEWER**

This PR includes changes to `resources.go` to allow for the passing of a `context.Background()` to `p.Meta` and `p.Refresh`.

This PR fixes #https://github.com/pulumi/pulumi-azure/issues/1655.

---

This PR was generated via `$ upgrade-provider pulumi/pulumi-azure --kind=bridge`.

---

- Upgrading pulumi-terraform-bridge from v3.71.0 to v3.72.0.


t0yv0:
That seems really odd. When is this code running? Is it running on provider startup, it's possibly slowing down provider startup then for no good reason. This likely should be `context.TODO()` for now until we figure it out...

provider/resources.go
```diff
@@ -3477,7 +3478,7 @@ func Provider() tfbridge.ProviderInfo {
 					res.Fields[azureLocation] = &tfbridge.SchemaInfo{
 						Name: azureLocation,
 						Default: &tfbridge.DefaultInfo{
-							From: defaultAzureLocation(p, rgRegionMap),
+							From: defaultAzureLocation(ctx, p, rgRegionMap),
```
iwahbe, line 3481: It seems like `From` should be changed to better support this type of behavior. A global `ctx` like this isn't really the best `ctx` here. `From` should be of type:

```go
func(context.Context, *tfbridge.PulumiResource) (interface{}, error)
```

iwahbe (approved):
I think it would be slightly clearer to have the `ctx` originate from `defaultAzureLocation`, but since `ctx` is effetely immutable it doesn't matter. Feel free to leave as it.

LGTM

iwahbe:
> That seems really odd. When is this code running? Is it running on provider startup, it's possibly slowing down provider startup then for no good reason. This likely should be `context.TODO()` for now until we figure it out...

`defaultAzureLocation` just returns a closure. It shouldn't negatively impact startup time:

https://github.com/pulumi/pulumi-azure/blob/6c88b7f781911804ee33c08791020982af1ea6d7/provider/resources.go#L3491-L3495

The code is run as part of default application for top level fields named `location`:

https://github.com/pulumi/pulumi-azure/blob/6c88b7f781911804ee33c08791020982af1ea6d7/provider/resources.go#L3466-L3486

PR #1681, opened 2024-02-05
secrets are leaking to plain text in stacks

ahmed-wajid:


ahmed-wajid:
To fix #1681 Haven't updated other SDKs since we are using only python at the moment.

ahmed-wajid:
@pierskarsenbarg  Could you please give us more information on how this can be fixed if we cannot directly change language SDKs?

PR #1686, opened 2024-02-06
Mark sensitive config settings

### LINKED ISSUE #1679: Sensitive data is exposed using azure classic provider
### What happened?

When I initialize the Azure classic provider with Azure sensitive data i.e. client id/client secret to deploy some Azure resources, the sensitive data is not encrypted in the Pulumi stack and there it's exposed in plain text.

We have used Azure Classic instead of Azure native library because of this bug described here: 
https://github.com/pulumi/pulumi-azure-native/issues/684

Can be this solved with another version of Azure Classic provider as the Azure Native is built or what's your recommendation to mitigate this high-security issue? 

If there is not an easy fix to solve this, can you provide me some references about how to migrate from Azure Classic to Azure Native v1 and then to Azure Native v2  WITHOUT deleting the existing resources deployed in a production environment? 

### Example

The invoked object for Azure Classic provider: 
image

Azure Classic provider implementation:
image
Here, the sensitive data is NOT encrypted in pulumi stack. 

I would expect to be implemented in the same way as the below
Azure Native provider implementation: 
image
Here, the sensitive data is encrypted in pulumi stack. 

### Output of `pulumi about`

The id and secret of the client are exposed in plain text in the Pulumi stack for the Azure Classic Provider input. 

### Additional context

For azure native v1 to v2 migration, we have this documentation:
https://www.pulumi.com/registry/packages/azure-native/from-v1-to-v2/

What about Classic to Native migration? It this possible?

BR, Gabriel

### Contributing

t0yv0:
As https://github.com/pulumi/pulumi-azure/issues/1679 demonstrates the provider has some authentication-related configuration settings that are not marked as sensitive in the upstream schema, and therefore are not treated as secret by Pulumi, but instead plaintexted in the state file. This change reconfigures the bridged provider to treat the properties as secret. 

Fixes #1679 

The suggested list of properties includes:

- auxillary_tenant_ids
- client_certificate
- client_certificate_password
- client_certificate_path
- client_id
- client_id_file_path
- client_secret
- client_secret_file_path
- oidc_request_token
- oidc_token
- oidc_token_file_path
- subscription_id
- tenant_id

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -548,6 +548,7 @@ func Provider() tfbridge.ProviderInfo {
 	// Adjust the defaults if running in Azure Cloud Shell.
 	// Environment variables still take preference, e.g. USE_MSI=false disables the MSI endpoint.
 	cloudShell := detectCloudShell()
+	yes := true
 
 	prov := tfbridge.ProviderInfo{
 		P:                p,
@@ -567,6 +568,7 @@ func Provider() tfbridge.ProviderInfo {
 					Value:   cloudShell.subscriptionID,
 					EnvVars: []string{"ARM_SUBSCRIPTION_ID"},
 				},
+				Secret: &yes,
 			},
 			"environment": {
 				Default: &tfbridge.DefaultInfo{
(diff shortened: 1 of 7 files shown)


pierskarsenbarg (approved):
LGTM

PR #1690, opened 2024-02-07
Upgrade terraform provider azurerm to v3.90.0

t0yv0:
This is a mostly routine update to v3.90.0 of upstream provider, but due to a dependency shift it required revisiting some of the patches and could not proceed automatically via upgrade-provider, notably:

```
github.com/hashicorp/go-azure-sdk v0.20240124.1115501
github.com/hashicorp/go-azure-sdk/sdk v0.20240201.1064937
```

Fixes https://github.com/pulumi/pulumi-azure/issues/1672


t0yv0:
I'm adding some temporary edits to the auto-generated lint.yml that essentially disables cache fixing https://github.com/pulumi/pulumi-azure/issues/1691 - I think subsequent nightly ci-mgmt jobs will revert the change, bringing the cache back in but it should work at that point hopefully.

guineveresaenger (approved):

PR #1729, opened 2024-02-19
Update README.md - fix dotnet example formatting

Meir017:


iwahbe (approved):
Thanks for catching this @Meir017

Issue #1994, opened 2024-05-01
PANIC in documentation


t0yv0:
### What happened?

There are panics rendered into documentation. 

Similarly to https://github.com/pulumi/pulumi-random/pull/890 these could be fixable by upgrading aux provider deps in .ci-mgmt.yml

### Example

Running `git grep panic` produces output but should not:

```
[...]

PR #1996, opened 2024-05-02
Bump all Pulumi plugin dependencies to latest

thomas11:
Main reason is to see if it helps with #1994.


iwahbe:
> How do we re-generate the workflows and makefile? cc: @guineveresaenger

Run `make ci-gmtm`.

iwahbe:
Hey @thomas11. This broke workflow updates, since this changes the schema. You need to run `make ci-mgmt tfgen` for this to be safe. See https://github.com/pulumi/pulumi-azure/pull/2059 as an example of a broken workflow update.

PR #2057, opened 2024-05-22
Improve patches

danielrbradley:
- Break large diffs apart into logical changes.
- Improve context in message of each patch.
- Remove a few obsolete patches.


thomas11:
For posteriority, it might be useful to describe a bit why the removed patches can be removed and how you determined that.

thomas11 (approved):

PR #2164, opened 2024-07-01
Enable TypedDict input types for the Python SDK

julienp:
Enable TypedDict input types for the Python SDK. The only manual change is to set the InputTypes flag in the python language features of the schema. This is done here https://github.com/pulumi/pulumi-azure/commit/e0ccec8323adde4f030c3daf312386ebb387ba9f. This flag is currently opt in, but will default to classes-and-dicts in the future.

Epic https://github.com/pulumi/pulumi/issues/12689


thomas11:
LGTM

PR #2217, opened 2024-07-11
Re-enable EventGrid test

thomas11:
Resolves #1036

Discovered this while working on something else related to EventGrid. Not sure what the original problem was but this change works for me locally. We'll see about CI shortly.

PR #2225, opened 2024-07-15
Migrate from @azure/ms-rest-nodeauth to @azure/identity as the former is deprecated and has a CVE

### LINKED ISSUE #1440: [CVE-2023-45857] Axios Cross-Site Request Forgery Vulnerability in transitive dependency
### What happened?

Github currently reports a vulnerability for Axios on the versions `>= 0.8.1, < 1.6.0`. Axios is used in [`adal-node`](https://github.com/Azure/ms-rest-nodeauth/issues/146), a transitive dependency of `azure/ms-rest-nodeauth`, which again is used by `pulumi/azure`. 

When investigating the dependency chain, I noticed that `adal-node` is no longer maintained. `azure/ms-rest-nodeauth` is also not longer actively developed and is not planning to removed `adal-node`. See https://github.com/Azure/ms-rest-nodeauth/issues/128. Instead, they suggest to migrate to `@azure/identity`.

This is blocking us from addressing the initial Axios security warning. Do you have plans to migrate away from `azure/ms-rest-nodeauth`?

### Example

N/A

### Output of `pulumi about`

N/A

### Additional context

Affects `@pulumi/azure` v5.55.0 

### Contributing

thomas11:
Resolves #1440

The NodeJS SDK has some hand-written mixins that haven't been updated in a long time. This PR replaces the deprecated @azure/ms-rest-nodeauth package, that has a CVE open (see #1440), with the current @azure/identity package.

The previous PR #2217 restored some test coverage for this change, although not all auth methods are covered. CI uses Service Principal with client secret, and I ran it locally using CLI auth.


sdk/nodejs/core/zMixins.ts
```diff
@@ -22,23 +22,26 @@ import * as utilities from "../utilities";
...
+    } else if (clientId && tenantId) {
+        if (clientSecret) {
+            tokenCredential = new identity.ClientSecretCredential(tenantId, clientId, clientSecret);
+        } else if (clientCertPath) {
+            tokenCredential = new identity.ClientCertificateCredential(tenantId, clientId, { certificatePath: clientCertPath, certificatePassword: clientCertPassword });
+        }
     }
 
-    return credentials;
+    console.log("Using Azure credentials:", tokenCredential.constructor.name);
```
mikhailshilkov, line 44: Double checking: is this added on purpose? Where does it pop up?

mikhailshilkov (approved):
LGTM if you trust this is well tested

iwahbe (approved):
Beyond the log line that @mikhailshilkov mentioned, this looks good to me.

sdk/nodejs/core/zMixins.ts
thomas11: good catch, leftover from development

PR #2505, opened 2024-10-01
Do not depend on specific sdk

t0yv0:
Re: https://github.com/pulumi/ci-mgmt/issues/1091


mjeffryes (approved):
Looks correct for the 1.133.0 version of codegen used by bridged repos

PR #2810, opened 2024-12-17
Remove `make upstream` from `preTest`

blampe:
After https://github.com/pulumi/ci-mgmt/pull/1151 our `make` targets should now all correctly call `make upstream` if they need it.

This updates `.ci-mgmt.yaml` and re-runs `make ci-mgmt`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -91,9 +91,6 @@ jobs:
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
         version: v2.5.0
-    - name: make upstream
-      run: |
-        make upstream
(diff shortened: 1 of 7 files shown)


.github/workflows/test.yml
```diff
@@ -91,9 +91,6 @@ jobs:
       with:
         token: ${{ secrets.GITHUB_TOKEN }}
         version: v2.5.0
-    - name: make upstream
```
t0yv0, line 94: Not having this in a separate step mixes up the logs a bit. But hopefully it never fails now.

t0yv0 (approved):

Issue #3030, opened 2025-02-19
Name length rule for NewKubernetesCluster is more strict than in azure


alex92ch:
### Describe what happened

Defining the cluster name using the unique `name` of the Pulumi resource is restricted to 24 characters, even though Azure allows up to 63 characters. Defining the name using the KubernetesClusterArgs.Name works, but I would like to use the unique suffixes as well.

<img width="728" alt="Image" src="https://github.com/user-attachments/assets/3c390502-9ac5-41a2-bb84-f595243926de" />

### Sample program

```
	aks, err := containerservice.NewKubernetesCluster(ctx, "aks-aviato-abt-container-cluster-", &containerservice.KubernetesClusterArgs{
		ResourceGroupName:       args.ResourceGroupName,
		WorkloadIdentityEnabled: pulumi.Bool(true),
		AzureRbacEnabled: pulumi.Bool(false),
		TenantId:         args.ClusterAadTenantID,
	        ...
                ...
         },
```

### Affected Resource(s)

NewKubernetesCluster

### Output of `pulumi about`

CLI          
Version      3.150.0
Go Version   go1.23.6
Go Compiler  gc

Plugins
KIND      NAME          VERSION
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azure-native  2.85.0
resource  azuread       5.53.5
resource  datadog       4.42.0
language  go            3.150.0
resource  random        4.17.0

Host     
OS       darwin
Version  15.3.1
Arch     arm64

This project is written in go: executable='/opt/homebrew/bin/go' version='go version go1.24.0 darwin/arm64'

### Contributing

Vote on this issue by adding a 👍 reaction. 
To contribute a fix for this issue, leave a comment (and link to your pull request, if you've opened one already).

rquitales:
[...]

PR #3031, opened 2025-02-19
fix: enable up to 63 chars for Kubernetes autonaming

rquitales:
This PR enables generating a physical name up to 63 characters for `KubernetesCluster` resources. The default currently is 24 characters.

Closes: #3030

### DIFF
--- a/provider/resources.go
+++ b/provider/resources.go
@@ -930,6 +930,16 @@ func Provider() tfbridge.ProviderInfo {
 					}
 					return pm, nil
 				},
+				Fields: map[string]*tfbridge.SchemaInfo{
+					// Use default autonaming options, but set max length to 63 as allowed by Azure.
+					// https://github.com/pulumi/pulumi-azure/issues/3030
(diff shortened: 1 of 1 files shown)


provider/resources.go
```diff
@@ -925,6 +925,15 @@ func Provider() tfbridge.ProviderInfo {
 					}
 					return pm, nil
 				},
+				Fields: map[string]*tfbridge.SchemaInfo{
+					// Use default autonaming options, but set max length to 64 as allowed by Azure.
+					azureName: tfbridge.AutoNameWithCustomOptions(azureName, tfbridge.AutoNameOptions{
+						Separator: "-",
+						Maxlen:    64,
```
thomas11, line 932: According to #3030 is should be 63, no?

thomas11 (changes requested):

provider/resources.go
rquitales: Good catch!

thomas11 (approved):

PR #3243, opened 2025-08-13
provider: support for loading the environment from the metadata host

tombuildsstuff:
This PR adds support for sourcing the environment from the metadata host.

If a `metadataHost` (or `ARM_METADATA_HOSTNAME` is specified) then the environment defined in `environment`/`ARM_ENVIRONMENT` is ignored and the environment configuration is instead loaded from the specified endpoint.


EronWright (approved):
Looks good to me!

I took a bit of time to understand why the `preConfigureCallback` exists here, since the underlying azurerm provider has native support.  I believe the reason is purely for config validation such that a Pulumi-specific message may be produced:
https://github.com/pulumi/pulumi-azure/pull/1130

guineveresaenger (approved):
This looks good to me!

The callback exists so that we are able to error early and let the user know about a misconfiguration.

PR #3543, opened 2026-06-02
chore: improve credential resolving (fixes #3471)

### LINKED ISSUE #3471: Using the provider in the Pulumi Kubernetes Operator with AKS Workload Identity only works with ARM_OIDC_TOKEN
### Describe what happened

I started using the Pulumi Kubernetes Operator in combination with Pulumi apps that use the azure-classic provider. The operator is running on an AKS cluster (k8s v1.35). The Service Account for the workspace pod has a workload identity with sufficient role assignments for the corresponding Pulumi app.

I'm mapping the Environment Variables that the workload identity webhooks mutates into the pod to ARM variables like this and the only way I can get the Authentication to work is to define the following init container for the workspace pod in my `Stack` CR:

```yaml
spec:
  config:
    azure:resourceProviderRegistrations: none
    azure:storageUseAzuread: true
    azure:subscriptionId: <redacted>
    azure:useOidc: true
  workspaceTemplate:
    spec:
      image: <my-comapny-docker-hub-mirror>/pulumi/pulumi-nodejs:3.228.0
      podTemplate:
        metadata:
          labels:
            azure.workload.identity/use: "true"
        spec:
          containers: []
          initContainers:
            - name: setup-env
              image: <my-comapny-docker-hub-mirror>/busybox
              # ARM_OIDC_TOKEN_FILE_PATH does not work, so we have to read the token ourselves.
              command:
                - sh
                - -c
                - |
                  echo "ARM_CLIENT_ID=${AZURE_CLIENT_ID}" >> $PULUMI_ENV
                  echo "ARM_TENANT_ID=${AZURE_TENANT_ID}" >> $PULUMI_ENV
                  echo "ARM_OIDC_TOKEN=$(cat /var/run/secrets/azure/tokens/azure-identity-token)" >> $PULUMI_ENV
              volumeMounts:
                - name: share
                  mountPath: /share
```

This feels pretty counter-intuitive, as there are other more idiomatic ways to use workload identity. What I tried before finding said solution:

1. Setting `ARM_OIDC_TOKEN_FILE_PATH` to  `/var/run/secrets/azure/tokens/azure-identity-token` or `${AZURE_FEDERATED_TOKEN_FILE}`
2. Using `ARM_USE_AKS_WORKLOAD_IDENTITY=true`  instead of using the `ARM_OIDC_...` variables
3. Using `azure:...` config keys for all of that instead of using env variables

None of these ways work at always result in the same error:

```
2026-04-09T11:19:47.795Z	WARN	pulumi	error: rpc error: code = Unknown desc = failed to load application credentials:
2026-04-09T11:19:47.796Z	WARN	pulumi	Details: could not configure AzureCli Authorizer: could not parse Azure CLI version: launching Azure CLI: exec: "az": executable file not found in $PATH
2026-04-09T11:19:47.796Z	WARN	pulumi	
2026-04-09T11:19:47.796Z	WARN	pulumi		Please make sure you have signed in via 'az login' or configured another authentication method.
2026-04-09T11:19:47.796Z	WARN	pulumi	
2026-04-09T11:19:47.796Z	WARN	pulumi		See https://www.pulumi.com/registry/packages/azure/installation-configuration/ for more information.
```

Am I doing anything wrong here or could this be a bug in the provider?

### Sample program

The content of the program does not matter, any Pulumi app using `@pulumi/azure` will be sufficient.

### Log output

_No response_

### Affected Resource(s)

_No response_

### Output of `pulumi about`

* Pulumi Kubernetes Operator v2.7.0 
* Workspace pod image: pulumi-nodejs:3.228.0
* azure-classic provider: v6.32.0
  * Can't update to a newer version because of: https://github.com/hashicorp/terraform-provider-azurerm/issues/31730  

### Additional context

_No response_

### Contributing

jalmeroth:
### Problem

The `preConfigureCallback` in `provider/resources.go` performs credential validation before the upstream Terraform provider configures itself. This validation was missing support for:

- `ARM_USE_AKS_WORKLOAD_IDENTITY` / `azure:useAksWorkloadIdentity`
- `ARM_OIDC_TOKEN_FILE_PATH` / `azure:oidcTokenFilePath`
- `AZURE_FEDERATED_TOKEN_FILE` (injected by AKS workload identity webhook)

When using AKS Workload Identity, the pre-validation fell through to Azure CLI auth, failed with "az: executable file not found", and never gave the underlying TF provider a chance to authenticate.

### Fix

Extract credential resolution into a `resolveCredentials` function that mirrors the upstream terraform-provider-azurerm logic:

- `useAksWorkloadIdentity` implies `useOidc = true`
- Falls back to `AZURE_CLIENT_ID` / `AZURE_TENANT_ID` env vars when AKS workload identity is enabled
- Resolves OIDC token from: direct value > file path (`ARM_OIDC_TOKEN_FILE_PATH`) > AKS token file (`AZURE_FEDERATED_TOKEN_FILE`)

### Testing

Added unit tests for `resolveCredentials` covering:
- AKS workload identity reads env vars and token file
- Explicit config takes precedence over env vars
- `oidcTokenFilePath` reads from file
- Direct `oidcToken` takes precedence over file
- Missing token file returns clear error
- OIDC stays disabled when not configured

fixes #3471

### DIFF
--- a/provider/preconfigure_test.go
+++ b/provider/preconfigure_test.go
@@ -32,7 +32,7 @@ func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
 		assert.True(t, creds.useOIDC)
 	})
 
-	t.Run("explicit config takes precedence over AZURE_* env vars", func(t *testing.T) {
+	t.Run("AZURE_* env vars take precedence over config when AKS WI enabled", func(t *testing.T) {
 		t.Setenv("AZURE_CLIENT_ID", "aks-client-id")
 		t.Setenv("AZURE_TENANT_ID", "aks-tenant-id")
 		t.Setenv("AZURE_FEDERATED_TOKEN_FILE", tokenFile)
@@ -45,8 +45,8 @@ func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
 
 		creds, err := resolveCredentials(vars)
 		require.NoError(t, err)
-		assert.Equal(t, "explicit-client", creds.clientID)
-		assert.Equal(t, "explicit-tenant", creds.tenantID)
+		assert.Equal(t, "aks-client-id", creds.clientID)
+		assert.Equal(t, "aks-tenant-id", creds.tenantID)
 	})
 
 	t.Run("oidcTokenFilePath reads token from file", func(t *testing.T) {
--- a/provider/preconfigure_test.go
+++ b/provider/preconfigure_test.go
@@ -5,9 +5,10 @@ import (
 	"path/filepath"
 	"testing"
 
-	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 	"github.com/stretchr/testify/assert"
 	"github.com/stretchr/testify/require"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
 )
 
 func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
@@ -55,9 +56,9 @@ func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
 		t.Setenv("AZURE_FEDERATED_TOKEN_FILE", "")
 
 		vars := resource.PropertyMap{
-			"useOidc":          resource.NewBoolProperty(true),
-			"clientId":         resource.NewStringProperty("test-client"),
-			"tenantId":         resource.NewStringProperty("test-tenant"),
+			"useOidc":           resource.NewBoolProperty(true),
+			"clientId":          resource.NewStringProperty("test-client"),
+			"tenantId":          resource.NewStringProperty("test-tenant"),
 			"oidcTokenFilePath": resource.NewStringProperty(tokenFile),
 		}
 
--- a/provider/preconfigure_test.go
+++ b/provider/preconfigure_test.go
@@ -1,0 +1,107 @@
+package provider
+
+import (
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/pulumi/pulumi/sdk/v3/go/common/resource"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
+	tokenContent := "my-federated-token"
+	tokenFile := filepath.Join(t.TempDir(), "azure-identity-token")
+	require.NoError(t, os.WriteFile(tokenFile, []byte("  "+tokenContent+"  \n"), 0o600))
+
+	t.Run("reads AZURE_* env vars and token file", func(t *testing.T) {
+		t.Setenv("AZURE_CLIENT_ID", "aks-client-id")
+		t.Setenv("AZURE_TENANT_ID", "aks-tenant-id")
+		t.Setenv("AZURE_FEDERATED_TOKEN_FILE", tokenFile)
+
+		vars := resource.PropertyMap{
+			"useAksWorkloadIdentity": resource.NewBoolProperty(true),
+		}
+
+		creds, err := resolveCredentials(vars)
+		require.NoError(t, err)
+		assert.Equal(t, "aks-client-id", creds.clientID)
+		assert.Equal(t, "aks-tenant-id", creds.tenantID)
+		assert.Equal(t, tokenContent, creds.oidcToken)
+		assert.True(t, creds.useOIDC)
+	})
+
+	t.Run("explicit config takes precedence over AZURE_* env vars", func(t *testing.T) {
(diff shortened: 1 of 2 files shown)


provider/preconfigure_test.go
```diff
@@ -0,0 +1,107 @@
...
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestResolveCredentials_AksWorkloadIdentity(t *testing.T) {
+	tokenContent := "my-federated-token"
+	tokenFile := filepath.Join(t.TempDir(), "azure-identity-token")
+	require.NoError(t, os.WriteFile(tokenFile, []byte("  "+tokenContent+"  \n"), 0o600))
+
+	t.Run("reads AZURE_* env vars and token file", func(t *testing.T) {
```
guineveresaenger, line 19: This test currently fails as follows:

```
=== FAIL: . TestResolveCredentials_AksWorkloadIdentity/reads_AZURE_*_env_vars_and_token_file (0.00s)
    preconfigure_test.go:29: 
        	Error Trace:	/home/runner/work/pulumi-azure/pulumi-azure/provider/preconfigure_test.go:29
        	Error:      	Not equal: 
        	            	expected: "aks-client-id"
        	            	actual  : "***"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-aks-client-id
        	            	+***
        	Test:       	TestResolveCredentials_AksWorkloadIdentity/reads_AZURE_*_env_vars_and_token_file
    preconfigure_test.go:30: 
        	Error Trace:	/home/runner/work/pulumi-azure/pulumi-azure/provider/preconfigure_test.go:30
        	Error:      	Not equal: 
        	            	expected: "aks-tenant-id"
        	            	actual  : "***"
        	            	
        	            	Diff:
        	            	--- Expected
        	            	+++ Actual
        	            	@@ -1 +1 @@
        	            	-aks-tenant-id
        	            	+***
        	Test:       	TestResolveCredentials_AksWorkloadIdentity/reads_AZURE_*_env_vars_and_token_file
    --- FAIL: TestResolveCredentials_AksWorkloadIdentity/reads_AZURE_*_env_vars_and_token_file (0.00s)

=== FAIL: . TestResolveCredentials_AksWorkloadIdentity (0.00s)
```

provider/resources.go
```diff
@@ -455,6 +455,60 @@ func detectCloudShell() cloudShellProfile {
...
+	useAksWorkloadIdentity := tfbridge.ConfigBoolValue(vars, "useAksWorkloadIdentity", []string{"ARM_USE_AKS_WORKLOAD_IDENTITY"})
+	useOIDC := tfbridge.ConfigBoolValue(vars, "useOidc", []string{"ARM_USE_OIDC"}) || useAksWorkloadIdentity
+
+	clientID := tfbridge.ConfigStringValue(vars, "clientId", []string{"ARM_CLIENT_ID"})
+	tenantID := tfbridge.ConfigStringValue(vars, "tenantId", []string{"ARM_TENANT_ID"})
+
+	// When using AKS Workload Identity, read client/tenant from the environment variables
+	// injected by the AKS workload identity webhook.
+	if useAksWorkloadIdentity {
+		if clientID == "" {
```
guineveresaenger: I think this is backwards from what upstream does: https://github.com/hashicorp/terraform-provider-azurerm/blob/f39d4c08c93aabdaf6191cae0520da1e008e7d8a/internal/provider/helpers.go#L97 

checks `AZURE_CLIENT_ID` first, and asks to reconcile if `ARM_CLIENT_ID` is set. We don't want the reconciliation error but we do want `AZURE_CLIENT_ID` to win.

guineveresaenger (changes requested):
Hi @jalmeroth, thank you for your pull request.

A preliminary test run surfaces that order of reading `AZURE*` env vars and `ARM_` env vars matters. Thank you for writing the test that caught that!

provider/preconfigure_test.go
guineveresaenger: This is because we're pre-setting some credentials in our CI environment. The test should probably clear those before running.

jalmeroth:
Thanks for your feedback @guineveresaenger! The latest change should address the precedence issue:  AZURE_CLIENT_ID now wins over ARM_CLIENT_ID when AKS Workload Identity is enabled.

jalmeroth:
I'm sorry, I broke the linting. Should be fixed now.

guineveresaenger:
thank you @jalmeroth - I often just patch up the linter myself; it tends to be a bit whiny, so I didn't say anything 😅 

Anyway everything else is green so I think we're g2g!

guineveresaenger (approved):

